PR #11, opened 2020-10-26
Attempt to produce correct valid obs number when qc=0

CoryMartin-NOAA:
checks to find where inverse_observation_error is not zero and adds that to the boolean and for the indices.

Has not been tested yet.

@kevindougherty-noaa let's try it on ch 7 for ABI and compare to the radmon site.
ch 7, qc =0, errcheck=True.


CoryMartin-NOAA:
@kevindougherty-noaa I had a typo, that hopefully has been fixed, errcheck should only count if qcflag == 0.

PR #19, opened 2020-12-22
Feature/pr practice

kevindougherty-noaa:
This was a practice pull request to get me familiar with the process.

I addressed the corrections requested in the last pull request besides the pressure issue which I will address in a different pull request.

### DIFF
--- a/pyGSI/diags.py
+++ b/pyGSI/diags.py
@@ -143,7 +143,6 @@ class Conventional(GSIdiag):
         super().__init__(path)
 
         self.read_conv_obs()
-        self.variable
 
     def __str__(self):
         return "Conventional object"
--- a/pyGSI/diags.py
+++ b/pyGSI/diags.py
@@ -13,6 +13,7 @@ class GSIdiag:
         """
 
         self.path = path
+        self.variable = path.split('/')[-1].split('.')[0].split('_')[2]
 
     def __len__(self):
         return len(self.lats)
@@ -28,15 +29,13 @@ class GSIdiag:
         ftype = self.path.split('/')[-1].split('.')[0].split('_')[-1]
 
         if dtype == 'conv':
-            variable = self.path.split('/')[-1].split('.')[0].split('_')[2]
-
             metadata = {'Diag_type': dtype,
-                        'Variable': variable,
+                        'Variable': self.variable,
                         'Date': date,
                         'File_type': ftype
                         }
         else:
-            satellite = self.path.split('/')[-1].split('.')[0].split('_')[2]
+            satellite = self.variable
 
             metadata = {'Diag_type': dtype,
                         'Satellite': satellite,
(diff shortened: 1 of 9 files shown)


pyGSI/diags.py
```diff
@@ -144,6 +143,7 @@ def __init__(self, path):
         super().__init__(path)
 
         self.read_conv_obs()
+        self.variable
```
CoryMartin-NOAA, line 146: This line is confusing to me?

CoryMartin-NOAA (dismissed):
One quick question/clarification, otherwise looks good!

scripts/create_conv_netcdf.py
```diff
@@ -145,10 +145,10 @@ def create_netcdf(YAML):
 else:
     nprocs = 1
 
-YAML = myargs.yaml
+input_yaml = myargs.yaml
 outdir = myargs.outdir
 
-file = open(YAML)
+file = open(input_yaml)
 parsed_yaml_file = yaml.load(file, Loader=yaml.FullLoader)
```
aerorahul, line 152: you should avoid this like the plague.
What happens if the file changes while it is open?  
The correct way to read from files is:
```
with open(input_yaml, 'r') as file:
    parsed_yaml_file = yaml.load(file, Loader=yaml.FullLoader)
```

scripts/create_sat_netcdf.py
```diff
@@ -57,10 +57,10 @@ def create_netcdf(YAML):
 else:
     nprocs = 1
 
-YAML = myargs.yaml
+input_yaml = myargs.yaml
 outdir = myargs.outdir
 
-file = open(YAML)
+file = open(input_yaml)
```
aerorahul, line 63: same comment as above.

scripts/mp_plot_conv_diags.py
```diff
@@ -123,17 +114,18 @@ def pool_handler(nprocs):
 else:
     nprocs = 1
 
-YAML = myargs.yaml
+input_yaml = myargs.yaml
 outdir = MyArgs.outdir
 
-file = open(YAML)
+file = open(input_yaml)
```
aerorahul, line 120: same as above.
not to mention, the file is never closed after opening.

scripts/create_conv_netcdf.py
CoryMartin-NOAA: This is good to know, I didn't understand why people always suggest the 'with open' syntax, is it because 'open' implies it is always open and is only read when you call the read method?

kevindougherty-noaa: Good to know. Thank you for letting me know

aerorahul: the `with` `open` line ensures that the method `close` is called on the `file` handle once you exit the scope, no matter the exception inside the `with` block. 

In general the `with` implements an `__enter__` and `__exit__` method that is executed as you enter and exit the scope of the code block.  It can be used elsewhere where you need to hold an object in memory temporarily.
I will not implement this if the file to be read is big.  The alternative would be to use the `try` `except` approach.

kevindougherty-noaa: A good explanation in addition to Rahul's: https://cmdlinetips.com/2016/01/opening-a-file-in-python-using-with-statement/

CoryMartin-NOAA (approved):
Ok the opens are fixed, and presumably that rogue line was fixed, so I approve now.

PR #20, opened 2020-12-30
Feature/filter station

kevindougherty-noaa:
Added an additional station id filter to the main method: get_data(). Now can input one or multiple stations and get the appropriate data back.

Added station filter to appropriate methods. 

Minor changes to pressure level data. Mostly made some changes to how it is read in and outputted. Will make more changes in a future pull request.

### DIFF
--- a/pyGSI/diags.py
+++ b/pyGSI/diags.py
@@ -307,7 +307,6 @@ class Conventional(GSIdiag):
                     return data
 
             else:
-                print(station_id)
                 idx = self.get_idx_conv(obsid, subtype, station_id, analysis_use)
 
                 if self.variable == 'uv':
--- a/pyGSI/diags.py
+++ b/pyGSI/diags.py
@@ -161,6 +161,7 @@ class Conventional(GSIdiag):
         self.press = f.variables['Pressure'][:]
         self.time = f.variables['Time'][:]
         self.anl_use = f.variables['Analysis_Use_Flag'][:]
+        self.stnid = f.variables['Station_ID'][:]
         try:
             self.stnelev = f.variables['Station_Elevation'][:]
         except:
@@ -179,7 +180,7 @@ class Conventional(GSIdiag):
 
         f.close()
 
-    def get_data(self, dtype, obsid=None, subtype=None, analysis_use=False, plvls=False):
+    def get_data(self, dtype, obsid=None, subtype=None, station_id=None, analysis_use=False, plvls=None):
         """
         Given parameters, get the data from a conventional diagnostic file
         INPUT:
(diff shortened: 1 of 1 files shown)


pyGSI/diags.py
```diff
@@ -189,23 +190,24 @@ def get_data(self, dtype, obsid=None, subtype=None, analysis_use=False, plvls=Fa
             optional:    
                 obsid        : observation measurement ID number; default=None
                 subtype      : observation measurement ID subtype number, default=None
+                station_id   : station id, default=None
                 analysis_use : if True, will return two sets of data: assimlated
                                (analysis_use_flag=1), and monitored (analysis_use
                                _flag=-1); default = False
-                plvls        : if True, will return a dictionary of data subsetting 
+                plvls        : if list given, will return a dictionary of data subsetting
```
CoryMartin-NOAA, line 197: Can you clarify in the doc string here? Like show an example. So presumably if I give you [0, 500, 1050], it would be two bins, [0-500,500-1050]? How is the dictionary defined, etc.?

kevindougherty-noaa: Do you mean to include this in the doc strings itself or just as an example right now?

CoryMartin-NOAA: In the docstring, so that a user can see how to use plvls.

kevindougherty-noaa: updated in newest commits

CoryMartin-NOAA (dismissed):
Looks good, I like the explicit 'P1-P2' dict key format.

pyGSI/diags.py
```diff
@@ -295,9 +307,10 @@ def get_data(self, dtype, obsid=None, subtype=None, analysis_use=False, plvls=Fa
                     return data
 
             else:
-                idx = self.get_idx_conv(obsid, subtype, analysis_use)
+                print(station_id)
```
aerorahul, line 310: do you need this print?

pyGSI/diags.py
```diff
@@ -307,22 +320,26 @@ def get_data(self, dtype, obsid=None, subtype=None, analysis_use=False, plvls=Fa
...
         OUTPUT:
             idx    : indices of the requested data in the file
         """
 
+        
         if not obsid or obsid == [None]:
             obsid = None
         if not subtype or subtype == [None]:
             subtype = None
+        if not station_id or station_id == [None]:
```
aerorahul, line 341: station_id is a list? 
I don't think you need these if not ... lines, since they are defaulted to `None` as `kwargs`.

aerorahul (approved):
looks good.
some minor comments.

pyGSI/diags.py
kevindougherty-noaa: This is mainly for the to address user YAML inputs. If a user leaves for example station id blank in the YAML or puts None and tries to run some of the examples in the repository, it would show as either [ ] or ["None"]. I added these lines just to be safe in case this did happen.

CoryMartin-NOAA (approved):

Issue #17, opened 2020-12-18
No absolute paths for data files


aerorahul:
Data paths in the yaml's should be relative paths to maximize portability.

PR #21, opened 2021-01-04
fixed absolute paths and tested batch scripts

kevindougherty-noaa:
This pull request is for issue #17. 

Fixed absolute paths in the batch shell scripts. 
Tested each batch script and addressed bugs (wrong variable and function names)

### DIFF
--- a/scripts/batch_conv_diags.sh
+++ b/scripts/batch_conv_diags.sh
@@ -8,11 +8,11 @@
 #SBATCH --mail-user=$LOGNAME@noaa.gov
 
 # PyGSIdir=path/to/PyGSI_Directory
-PyGSIdir=/scratch1/NCEPDEV/da/$LOGNAME/PyGSI
-OUTDIR=/scratch1/NCEPDEV/da/$LOGNAME/PyGSI/
+PyGSIdir=../
+OUTDIR=../
 YAML=$PyGSIdir/test_conv_yaml.yaml
 
 # load environment needed to run python scripts
 source $PyGSIdir/modulefiles/modulefile.PyGSI.hera.bash
 
-python $PyGSIdir/scripts/mp_plot_convDiags.py -n 20 -y $YAML -o $OUTDIR
+python $PyGSIdir/scripts/mp_plot_conv_diags.py -n 20 -y $YAML -o $OUTDIR
--- a/scripts/batch_sat_diags.sh
+++ b/scripts/batch_sat_diags.sh
@@ -1,4 +1,3 @@
-  
 #!/bin/bash
 #SBATCH -J plot_gsi_diags
 #SBATCH -A da-cpu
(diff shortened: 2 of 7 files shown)


scripts/batch_conv_diags.sh
```diff
@@ -8,11 +8,11 @@
 #SBATCH --mail-user=$LOGNAME@noaa.gov
 
 # PyGSIdir=path/to/PyGSI_Directory
-PyGSIdir=/scratch1/NCEPDEV/da/$LOGNAME/PyGSI
-OUTDIR=/scratch1/NCEPDEV/da/$LOGNAME/PyGSI/
+PyGSIdir=../
+OUTDIR=../
```
CoryMartin-NOAA, line 12: This only works if the scripts directory is the same directory you submit the job from. I think you want to use something like ``PyGSIdir=`dirname "$0"```. Note there are two extra `` because I can't figure out how to make a literal back tick in GitHub.

CoryMartin-NOAA: Also outdir should be somewhere else by default, even if it's just $PyGSIdir/out (may need a mkdir -p $OUTDIR)

scripts/batch_sat_diags.sh
```diff
@@ -9,11 +8,11 @@
 #SBATCH --mail-user=$LOGNAME@noaa.gov
 
 # PyGSIdir=path/to/PyGSI_Directory
-PyGSIdir=/scratch1/NCEPDEV/da/$LOGNAME/PyGSI
-OUTDIR=/scratch1/NCEPDEV/da/$LOGNAME/PyGSI/
+PyGSIdir=../
+OUTDIR=../
```
CoryMartin-NOAA, line 12: Same comments as above.

CoryMartin-NOAA (changes requested):
I have a couple of requested changes for this PR. 

Also, this doesn't totally address @aerorahul's issue he opened. I think what he was saying is the sample test YAML files have hard coded paths, and perhaps the better approach is to use relative paths here. I guess this is more of an issue with my convinfo2yaml scripts? We should probably move to this repo/modify those scripts.

scripts/batch_conv_diags.sh
aerorahul: There are several issues with this script:
1. `--mail-user=$LOGNAME@noaa.gov` should be eliminated.  A non-NOAA user should be able to run this.
2. This script sources a modulefile meant for hera.  We should eliminate that line or make it generic.  What if there are no modules?
3. Establish a `Data/` directory and then all paths are relative to that.  Relative paths should not be relative to current working directory, and having paths such as `../` are trip hazards.

CoryMartin-NOAA:
@aerorahul @kevindougherty-noaa I think what we need is a top-level wrapper/setup script that can copy/extract the GSI diags, set up the Data directory, create the YAML, and do whatever else it needs before submission.

My only issue is how best to make the 'module' generic. I don't particularly like the way the global workflow does it, but perhaps that's the way to go? (check for a path and if it exists, source a module file, else do nothing?)

Issue #22, opened 2021-01-04
Move convinfo2yaml and satinfo2yaml scripts to this repo


CoryMartin-NOAA:
https://github.com/corymartin-noaa/darth has scripts to generate YAML to run PyGSI in batch mode, move here and add a wrapper script so that the sample YAML files can be generated on the fly more easily and no hard coded paths.

PR #23, opened 2021-01-04
Add in convinfo2yaml and satinfo2yaml scripts from DARTH to PyGSI + sample convinfo and satinfo files

CoryMartin-NOAA:
Closes Issue #22.

Previously, the Python scripts convinfo2yaml.py and satinfo2yaml.py resided in DARTH, this moves them to within PyGSI. These scripts take an input data directory of GSI netCDF diagnostic files, and an input convinfo/satinfo file and produces YAML to use in the PyGSI batch scripts.

A future PR will add a wrapper script to run these Python scripts.

This PR also adds GFSv16 satinfo and convinfo files to parm/ for easy access/reference, but users may need to use a different convinfo/satinfo file as appropriate (for regional models for example)


CoryMartin-NOAA:
@aerorahul @kevindougherty-noaa this is the time to rip apart any issues with these hastily made scripts I haven't touched in months ;-)

ush/convinfo2yaml.py
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env python
```
aerorahul, line 1: I bet this doesn't pass pep8.

CoryMartin-NOAA: ha! almost certainly does not

CoryMartin-NOAA: This is a good opportunity for me to do that too.

ush/convinfo2yaml.py
```diff
@@ -0,0 +1,81 @@
...
+    cv.close()
+
+    # get list of conventional diagnostic files available
+    diagpath = config['diagdir'] + '/diag_conv_*_' + config['loop'] + '.' + config['cycle'] + '.nc4'
+    diagfiles = glob.glob(diagpath)
+
+    # initialize YAML dictionary for output
+    yamlout = { 'diagnostic': [] }
+    if config['loop'] == 'ges':
+        diagtype = 'O-F'
```
kevindougherty-noaa: Does this script only have the option of using O-F or O-A? Nothing for just Observation or h(x)? There are already a lot of inputs, but should this be something the user should input?

CoryMartin-NOAA: This is a good point, it should support that if the user wishes.

kevindougherty-noaa (approved):
Looks good to me. I will go ahead and approve

ush/convinfo2yaml.py
```diff
@@ -0,0 +1,104 @@
...
+                    default='omf')
+args = parser.parse_args()
+
+config = {}
+config['diagdir'] = args.diagdir.rstrip('/')
+config['cycle'] = args.cycle
+config['convinfo'] = args.convinfo
+config['yaml'] = args.yaml
+config['loop'] = args.loop
+config['variable'] = args.variable
```
aerorahul, line 102: why not just 
```
config = args
config['diagdir'] = config.diagdir.rstrip('/')
```

CoryMartin-NOAA:
@aerorahul I saw a comment come in via email but I can't find it here, argparse.parse_args returns a namespace (I think) and not a dictionary so it's not immediately clear to me how to add additional objects (aka keys to the dictionary) like is needed/done here

PR #26, opened 2021-01-05
Add in ozinfo file; add in ozinfo2yaml python script; 

CoryMartin-NOAA:
The GFSv16 ozinfo file is added to fix.

Added new Python script like the other two existing ones but for ozinfo2yaml.

Additionally, added a new command line argument to each Python script to optionally include monitored observations, levels, channels, etc. and changed the way the config dictionary is defined based off of the argparse args namespace.


ush/convinfo2yaml.py
```diff
@@ -91,14 +91,13 @@ def main(config):
 parser.add_argument('-v', '--variable', type=str,
                     help='read departures, obs, or H(x): omf | obs | hofx',
                     default='omf')
+parser.add_argument('-m', '--monitor', type=str,
+                    help='include monitored obs, yes or no?',
+                    default='no')
```
aerorahul, line 96: use `action=store_true` instead of yes/no, e.g.
```
parser.add_argument('--foobar', action='store_true')
```

aerorahul (changes requested):
mostly looks good. 
one comment on yes/no.  There is a way to provide that via action in argparse.

ush/convinfo2yaml.py
CoryMartin-NOAA: Done

ush/satinfo2yaml.py
```diff
@@ -87,14 +87,11 @@ def main(config):
...
 
-config = {}
-config['diagdir'] = args.diagdir.rstrip('/')
-config['cycle'] = args.cycle
-config['satinfo'] = args.satinfo
-config['yaml'] = args.yaml
-config['loop'] = args.loop
-config['variable'] = args.variable
+config = vars(args)
+config['diagdir'] = config['diagdir'].rstrip('/')
```
aerorahul, line 95: why `rstrip`? And even if there is a trailing `/`, so what?
When joining paths, you should use `os.path.join`

CoryMartin-NOAA: The trailing / was actually a problem, because I was doing "if X in Y" where Y is a list returned by glob.glob and there would be a mismatch of the path strings, but os.path.join seems like a much better option, so I'll try to rework with that.

ush/ozinfo2yaml.py
```diff
@@ -0,0 +1,102 @@
...
+parser.add_argument('-c', '--cycle', type=str,
+                    help='cycle YYYYMMDDHH', required=True)
+parser.add_argument('-i', '--ozinfo', type=str,
+                    help='path to GSI ozinfo file', required=True)
+parser.add_argument('-y', '--yaml', type=str,
+                    help='path to output YAML file', required=True)
+parser.add_argument('-l', '--loop', type=str,
+                    help='ges|anl default ges', default='ges')
+parser.add_argument('-v', '--variable', type=str,
+                    help='read departures, obs, or H(x): omf | obs | hofx',
```
aerorahul, line 93: use choices, e.g.
```
choices=['ges', 'anl']
```
That way, if anything else is passed, it throws an exception.
Same logic elsewhere where there are only a certain set of valid options

CoryMartin-NOAA: Perfect, thanks, I'm learning a lot more about argparse today!.

aerorahul: right about now, I suggest we move to `click` :-)

aerorahul (approved):
looks good.

PR #28, opened 2021-01-08
Feature/add ozone class

kevindougherty-noaa:
Added Ozone() class to diags.py.

Methods include read_ozone_obs(), get_data(), get_lat_lon(), and analysis_use_idx().

Depending on the type of diagnostic file, returns a dictionary of the total column ozone values and the different layer values. Only filter that can be applied is analysis_use_flag, which determines if the user would like the separated assimilated and monitored data.


pyGSI/diags.py
```diff
@@ -642,3 +638,240 @@ def get_lat_lon(self, channel=None, qcflag=None, errcheck=True):
...
+        
+        if analysis_use == False:
+            column_total_idx = np.where(self.ref_pressure == 0)
+            layer_idx = np.where(self.ref_pressure != 0)
+            
+            if layer_idx[0].size == 0:
+                return self.lats[column_total_idx], self.lons[column_total_idx]
+
+            else:
+                lats={}
```
aerorahul, line 809: move these 2 lines before the if-block.

pyGSI/diags.py
```diff
@@ -642,3 +638,240 @@ def get_lat_lon(self, channel=None, qcflag=None, errcheck=True):
...
+
+        self.lons = f.variables['Longitude'][:]
+        self.lats = f.variables['Latitude'][:]
+        self.ref_pressure = f.variables['Reference_Pressure'][:]
+        self.time = f.variables['Time'][:]
+        self.anl_use = f.variables['Analysis_Use_Flag'][:]
+        self.o = f.variables['Observation'][:]
+        self.omf = f.variables['Obs_Minus_Forecast_adjusted'][:]
+
+        f.close()
```
aerorahul, line 675: please wrap this in `with open ...` block.

pyGSI/diags.py
```diff
@@ -642,3 +638,240 @@ def get_lat_lon(self, channel=None, qcflag=None, errcheck=True):
...
+        
+
+    def get_data(self, dtype, analysis_use=False):
```
aerorahul, line 678: `dtype` is not a good choice here since it is a `numpy` object.

pyGSI/diags.py
```diff
@@ -642,3 +638,240 @@ def get_lat_lon(self, channel=None, qcflag=None, errcheck=True):
...
+
+                return data_dict
+
+            else:
+                for layer in self.ref_pressure:
+                    if layer != 0:
+                        idx = np.where(self.ref_pressure == layer)
+                        data = self.query_data_type(dtype, idx)
+
+                        data_dict[layer] = data
```
CoryMartin-NOAA, line 713: so the dictionary keys are float values + the string 'column_total'? Does that work? Is it ideal?

pyGSI/diags.py
```diff
@@ -642,3 +638,240 @@ def get_lat_lon(self, channel=None, qcflag=None, errcheck=True):
...
+            
+            if layer_idx[0].size == 0:
+                return self.lats[column_total_idx], self.lons[column_total_idx]
+
+            else:
+                lats={}
+                lons={}
+                for layer in self.ref_pressure:
+                    if layer != 0:
+                        idx = np.where(self.ref_pressure == layer)
```
CoryMartin-NOAA, line 813: like 95% sure that the lat/lon values will be the same for all layers, even total column, so this might be overkill/redundant. How do you do it for the radiances?

pyGSI/diags.py
```diff
@@ -642,3 +638,240 @@ def get_lat_lon(self, channel=None, qcflag=None, errcheck=True):
...
+                               _flag=-1); default = False
+
+        OUTPUT:
+            data   : requested data as a dictionary
+        """
+
+        data_dict = {}
+        
+        if analysis_use == False:
+
```
aerorahul, line 697: There are too many `if-else` constructs here, leading to too many repetitive code fragments.

aerorahul (changes requested):
see inline comments.

pyGSI/diags.py
kevindougherty-noaa: Other than that, I'm not too sure what would be best to decipher the different layers and the total column data.

CoryMartin-NOAA: ok as long as it works, I'm fine with it.

pyGSI/diags.py
kevindougherty-noaa: You are correct, was overthinking it. However, it is not just as straight forward as Radiance class. It's a combination of the both of them so trying to figure out the best way to do this. I'll work on it and address all comments in another commit.

pyGSI/diags.py
```diff
@@ -23,34 +23,34 @@ def get_metadata(self):
...
         return metadata
 
-    def query_data_type(self, dtype, idx):
+    def query_data_type(self, data_type, idx):
         """
         Query the data type being requested and returns
         the appropriate indexed data
         """
-        if dtype == 'O-F' or dtype == 'o-f':
+        if data_type == 'O-F' or data_type == 'o-f':
```
CoryMartin-NOAA, line 53: Can we make these something like (pseudo code) if data_type.ALLCAPS == 'O-F'? Same for below.

CoryMartin-NOAA: I think it's str.upper() or str.lower()? I think lowercase is probably preferred.

pyGSI/diags.py
```diff
@@ -152,40 +152,37 @@ def read_conv_obs(self):
...
-        self.lats = f.variables['Latitude'][:]
-        self.press = f.variables['Pressure'][:]
-        self.time = f.variables['Time'][:]
-        self.anl_use = f.variables['Analysis_Use_Flag'][:]
-        self.stnid = f.variables['Station_ID'][:]
-        try:
-            self.stnelev = f.variables['Station_Elevation'][:]
-        except:
-            self.modelelev = f.variables['Model_Elevation'][:]
+            if self.path.split('/')[-1].split('.')[0].split('_')[2] == 'uv':
```
CoryMartin-NOAA, line 169: if self.variable?

pyGSI/diags.py
```diff
@@ -23,34 +23,34 @@ def get_metadata(self):
         Grabs metadata from the diagnostic filename
         """
 
-        dtype = self.path.split('/')[-1].split('.')[0].split('_')[1]
+        data_type = self.path.split('/')[-1].split('.')[0].split('_')[1]
```
aerorahul, line 26: please use something understandable.
The first split is to get the filename. Use `os.path.basename`.
The second split is to remove the extension. Use `path.without_suffix` from `Pathlib`.
The third split is the one that actually splits the remainder of the string by `_` and get the 2nd item in the list.

aerorahul: Why `data_type`.  It is a `diag_type`

pyGSI/diags.py
kevindougherty-noaa: Okay. So if I do `if data_type.lower() == 'o-f'` should I include an "omf" option too?

pyGSI/diags.py
```diff
@@ -23,34 +23,34 @@ def get_metadata(self):
...
+            metadata = {'Diag_type': data_type,
                         'Variable': self.variable,
                         'Date': date,
                         'File_type': ftype
                         }
         else:
             satellite = self.variable
 
-            metadata = {'Diag_type': dtype,
+            metadata = {'Diag_type': data_type,
```
aerorahul, line 40: Keys don't need to have `_`.
You could name this: `Diagnostic` or `Diag Type`

pyGSI/diags.py
CoryMartin-NOAA: Maybe:
`if data_type.lower() in ['o-f', 'omf']:
`
?

aerorahul (changes requested):

pyGSI/diags.py
aerorahul: is `o-f` even a valid choice in the user argument that is passed?

pyGSI/diags.py
kevindougherty-noaa: Would this be alright?
`import os
from pathlib import Path
os.path.splitext(Path(<file>).stem)[0].split('_')[1]`

Path(<file>).stem grabs file name and removes .nc4.
os.path.splitext( )[0] grabs only the the first part of the filename
.split('_') grabs the item needed, in this case the 2nd item

pyGSI/diags.py
kevindougherty-noaa: Yes, however I will add omf/omb and oma if you think that is appropriate and what most users are comfortable with.

pyGSI/diags.py
```diff
@@ -23,34 +26,40 @@ def get_metadata(self):
         Grabs metadata from the diagnostic filename
         """
 
-        dtype = self.path.split('/')[-1].split('.')[0].split('_')[1]
-        str_date = self.path.split('/')[-1].split('.')[1]
+        diag_type = self.filename.split('_')[1]
+        str_date = os.path.basename(self.path).split('.')[1]
+        print(str_date)
```
CoryMartin-NOAA, line 31: Remove the print

pyGSI/diags.py
```diff
@@ -23,34 +26,40 @@ def get_metadata(self):
         Grabs metadata from the diagnostic filename
         """
 
-        dtype = self.path.split('/')[-1].split('.')[0].split('_')[1]
-        str_date = self.path.split('/')[-1].split('.')[1]
+        diag_type = self.filename.split('_')[1]
```
CoryMartin-NOAA, line 29: diag_type here means something different (conv?) than diag_type below (omb, oma, etc.). Can we have different names? Why is it ftype below the same as diag_type here?

CoryMartin-NOAA: obs type? perhaps?

pyGSI/diags.py
```diff
@@ -340,9 +347,9 @@ def get_idx_conv(self, obsid=None, subtype=None, station_id=None, analysis_use=F
         if not station_id or station_id == [None]:
             station_id = None
 
-        if analysis_use == False:
+        if not analysis_use:
             idxobs = self.o_type
-            valid_idx = np.full_like(idxobs, True, dtype=bool)
+            valid_idx = np.full_like(idxobs, True, diag_type=bool)
```
CoryMartin-NOAA, line 352: this find and replace was dangerous here, might be other places?

pyGSI/diags.py
aerorahul: shouldn't this be in the constructor?

kevindougherty-noaa: I see now. I believe the find and replace was more harmful than helpful. I'll go through each one and fix. This applies to other comments as well.

pyGSI/diags.py
```diff
@@ -23,34 +26,40 @@ def get_metadata(self):
...
         return metadata
 
-    def query_data_type(self, dtype, idx):
+    def query_diag_type(self, diag_type, idx):
         """
         Query the data type being requested and returns
         the appropriate indexed data
         """
-        if dtype == 'O-F' or dtype == 'o-f':
+        if diag_type.lower() is in ['o-f', 'omf', 'o-b', 'omb']:
```
aerorahul, line 62: I think this was asked earlier, I suggest we stick to `omf` and `oma`, and all lower case.  remove `o-f` and other variants.

pyGSI/diags.py
```diff
@@ -72,7 +81,7 @@ def query_data_type(self, dtype, idx):
                 data = self.o[idx]
                 return data
 
-        elif dtype == 'O-A' or dtype == 'o-a':
+        elif diag_type.lower() is in ['o-a', 'omb']:
```
aerorahul, line 84: o-a and omb are different quantities.

pyGSI/diags.py
```diff
@@ -81,7 +90,7 @@ def query_data_type(self, dtype, idx):
                 print('File type does not support O-A')
                 return None
 
-        elif dtype == 'H(x)' or dtype == 'h(x)' or dtype == 'hofx':
+        elif diag_type.lower() isin ['h(x)', 'hofx']:
```
aerorahul, line 93: abandon `h(x)` and stick with `hofx`

pyGSI/diags.py
```diff
@@ -93,36 +102,36 @@ def query_data_type(self, dtype, idx):
                 data = hx[idx]
                 return data
 
-        elif dtype == 'Water_Fraction' or dtype == 'water_fraction':
+        elif diag_type.lower() == 'water_fraction':
```
aerorahul, line 105: when you populate `diag_type` make it lower.  Remove `.lower()` from everywhere.

pyGSI/diags.py
```diff
@@ -93,36 +102,36 @@ def query_data_type(self, dtype, idx):
...
+            raise Exception(f'Unrecognizable data type: {diag_type}')
             return None
 
     def check_o_minus_a(self):
         """
         Checks if the diagnostic file is an analysis file.
         """
 
-        ftype = self.path.split('/')[-1].split('.')[0].split('_')[-1]
+        ftype = self.filename.split('_')[-1]
```
aerorahul, line 134: again, this should be in the constructor.

pyGSI/diags.py
kevindougherty-noaa: So for conventional vs. radiance vs. ozone, what would be a better variable for determining the type it is like in this method? Follow up, what would be better for like the omf, oma etc. I believe my nomenclature is limited and I'm still learning how to better name variables.

CoryMartin-NOAA: I think ob_type or maybe ob_class works for me, there are 3, conv, rad, oz.

pyGSI/diags.py
CoryMartin-NOAA: I'm torn on this, on the one hand, I see no harm in having multiple choices, but on the other, it's fine to be consistent and force the users to just use one choice.

pyGSI/diags.py
CoryMartin-NOAA: Yeah this is a typo here I think

pyGSI/diags.py
CoryMartin-NOAA: I agree with this one 100%

pyGSI/diags.py
aerorahul: you can have multiple inputs, but better to redefine once and use that everywhere.

pyGSI/diags.py
kevindougherty-noaa: Okay noted. Would diag_type still be good when referring to the input by the user in terms of omf, oma, observation, etc.?

pyGSI/diags.py
CoryMartin-NOAA: Yeah, perhaps instead of ftype, have two booleans, self.anl = True and self.ges = False, and vice versa?

aerorahul: `is_anl` and `is_ges`
Initialize both to `False` and then set one or the other to `True`.

pyGSI/diags.py
CoryMartin-NOAA: Perhaps for readability, something like this:
at the very top of the method
```
diag_type = diag_type.lower()
omf = True if diag_type in ['omf', o-f'] else False
oma = True if diag_type in ['oma', o-a'] else False
obs = True if diag_type in ['observation'] else False
hofx = True if diag_type in ['hofx'] else False

```
Then later on:

```
if omf:
  do something

if hofx:
  do something
```

pyGSI/diags.py
kevindougherty-noaa: It is a typo.

pyGSI/diags.py
kevindougherty-noaa: I like this. Should I also add the *_Fraction variables as well and keep it boolean?

CoryMartin-NOAA: yeah make everything boolean I think

pyGSI/diags.py
CoryMartin-NOAA: I think so, meaning 'type of diagnostic' to read/grab

pyGSI/diags.py
kevindougherty-noaa: so something like this? 

```
is_anl = False
is_ges = False

if self.ftype == 'ges':
    is_ges = True
else:
    is_anl = True
```

where `self.ftype` is defined in the __init__() constructor. Then I do not need the `check_o_minus_a()` method. and can just do (with doing what Cory mentioned above and making diag_types as boolean):

```
omf = True if diag_type in ['omf', 'o-f', 'omb', 'o-b'] else False

is_anl = False
is_ges = False

if self.ftype == 'ges':
    is_ges = True
else:
    is_anl = True

if omf and is_ges:
    do something
```

CoryMartin-NOAA: I would do
```
is_ges = True if self.ftype == 'ges' else False
is_anl = True if self.ftype == 'anl' else False
```
that way, in the rare event that it is neither ges nor anl, it doesn't assume anything.

kevindougherty-noaa: Got it, less lines of code as well.

pyGSI/diags.py
```diff
@@ -152,40 +161,38 @@ def read_conv_obs(self):
...
-            self.v_omf = f.variables['v_Obs_Minus_Forecast_adjusted'][:]
+                self.u_omf = f.variables['u_Obs_Minus_Forecast_adjusted'][:]
+                self.v_omf = f.variables['v_Obs_Minus_Forecast_adjusted'][:]
 
-        else:
-            self.o = f.variables['Observation'][:]
-            self.omf = f.variables['Obs_Minus_Forecast_adjusted'][:]
+            else:
+                self.o = f.variables['Observation'][:]
+                self.omf = f.variables['Obs_Minus_Forecast_adjusted'][:]
```
aerorahul, line 187: This is fine for now, but I would really like to use pandas dataframe here.

aerorahul: or xarray.

pyGSI/diags.py
```diff
@@ -152,40 +150,38 @@ def read_conv_obs(self):
...
-        f.close()
 
-    def get_data(self, dtype, obsid=None, subtype=None, station_id=None, analysis_use=False, plvls=None):
+    def get_data(self, diag_type, obsid=None, subtype=None, station_id=None, analysis_use=False, plvls=None):
         """
         Given parameters, get the data from a conventional diagnostic file
         INPUT:
             required:
-                dtype  : type of data to extract i.e. Observation, O-F, O-A, H(x)
+                diag_type  : type of data to extract i.e. Observation, O-F, O-A, H(x)
```
CoryMartin-NOAA, line 184: Maybe make the defaults here in the string 'observation, omf, Oma, hofx'?

pyGSI/diags.py
```diff
@@ -642,3 +638,240 @@ def get_lat_lon(self, channel=None, qcflag=None, errcheck=True):
...
+
+class Ozone(GSIdiag):
+
+    def __init__(self, path):
+        """
+        Initialize a conventional GSI diagnostic object
+        INPUT:
+            path   : path to ozone GSI diagnostic object
+        RESULT:
+            self   : GSIdiag object containing the path to extract data
```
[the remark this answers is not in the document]
CoryMartin-NOAA: Yeah I think the self : GSIdiag is suggested to be 'GSI ozone diag object' or something similar.

pyGSI/diags.py
```diff
@@ -642,3 +632,180 @@ def get_lat_lon(self, channel=None, qcflag=None, errcheck=True):
...
+        """
+        super().__init__(path)
+
+        self.read_ozone_obs()
+
+    def __str__(self):
+        return "Ozone object"
```
CoryMartin-NOAA, line 652: Maybe say 'ozone GSI diagnostic object' to be explicit?

CoryMartin-NOAA (commented):
Couple of minor comments still

CoryMartin-NOAA (approved):
I think I'm happy, will give sometime for @aerorahul to object to anything but I think we've successfully beaten this dead horse

pyGSI/diags.py
```diff
@@ -642,3 +636,180 @@ def get_lat_lon(self, channel=None, qcflag=None, errcheck=True):
...
+            self   : GSI ozone diag object containing the path to extract data
+        """
+        super().__init__(path)
+
+        self.read_ozone_obs()
+
+    def __str__(self):
+        return "Ozone GSI diagnostic object"
+
+    def read_ozone_obs(self):
```
aerorahul, line 658: name this method read_obs.
and the same in Radiance and Conventional

CoryMartin-NOAA: good catch, agreed. Should be consistent.

aerorahul: also, it is private, so it should be `_read_obs`

pyGSI/diags.py
```diff
@@ -13,44 +15,60 @@ def __init__(self, path):
...
-        if dtype == 'O-F' or dtype == 'o-f':
+                return u, v
+
+            else:
+                data = self.omf[idx]
+                return data
+
+        if oma and is_anl:
             if self.variable == 'uv':
                 u = self.u_omf[idx]
```
aerorahul, line 73: Is this correct?
Shouldn't this be `self.u_oma`

CoryMartin-NOAA: Pretty sure this is because of the way the GSI diag files are written. Regardless of if it is loop 1 or 3 (ges/anl), the variable is named 'Observation_minus_forecast', so the part of the code that actually interfaces with netCDF just names the object self.omf. To make it self.oma for oma, we would probably need to add an if/else into the part of the code that reads the variable from the netCDF file.

kevindougherty-noaa: Yes, I just double checked and no matter if the file is ges or anl, it is always 'Obs_Minus_Forecast_adjusted'. I can throw a conditional statement just so it is clear to future users.

CoryMartin-NOAA: Maybe just make them 'departures'? Or is that too confusing?
```
self.u_dep
self.v_dep
self.dep
```
Then here,
```
if oma and is_anl:
  return self.dep
```
?
I don't feel strongly in any direction but I feel like adding an extra if statement is just making it messier, especially since this is all 'internal' code that the average user won't see.

aerorahul (dismissed):

CoryMartin-NOAA (approved):

PR #29, opened 2021-01-19
Feature/update plotting binning

kevindougherty-noaa:
Updated plotting scripts and binning scripts to match variables in diags.py as well as files that create netcdf files


pyGSI/diags.py
```diff
@@ -13,7 +13,7 @@ def __init__(self, path):
         INPUT:
             path : path to GSI diagnostic object
         """
-
+
```
CoryMartin-NOAA, line 16: These extra lines are the only difference in this file? Can we checkout 'diags.py' from main and not make any changes to it in this PR?

pyGSI/netcdf_diags.py
```diff
@@ -511,45 +511,30 @@ def append_ncfile_uv(uv_stat_dict, binned_data, ncfilename):
...
+    
+    if omf:
+        metadata['Diag Type'] = 'OmF'
+    
+    if oma:
+        metadata['Diag Type'] = 'OmA'
+        
+        
+    if metadata['Obs Type'] == 'conv':
+        ncfilename = '{Obs Type}_{Variable}_{ObsID[0]}_{Diag Type}.nc'.format(
```
CoryMartin-NOAA, line 526: Rahul will tell you to use f strings haha

pyGSI/netcdf_diags.py
```diff
@@ -511,45 +511,30 @@ def append_ncfile_uv(uv_stat_dict, binned_data, ncfilename):
...
-            ncfilename = '{Diag_type}_{Variable}_{ObsID[0]}_OmA.nc'.format(
-                **metadata)
-        else:
-            ncfilename = '{Diag_type}_{Satellite}_Ch{Channel[0]}_OmA.nc'.format(
+    
+    omf = True if metadata['Diag Type'] in ['omf', 'o-f', 'omb', 'o-b'] else False
+    oma = True if metadata['Diag Type'] in ['oma', 'o-a'] else False
+    
+    if omf:
+        metadata['Diag Type'] = 'OmF'
```
CoryMartin-NOAA, line 519: Is there a better way to do this? Do we need all these extra lines and ifs?

pyGSI/plot_diags.py
```diff
@@ -89,12 +89,12 @@ def get_xlabel(metadata):
                    'windspeed': "Windspeed (m/s)"
                    }
 
-    if metadata['Diag_type'] == 'conv':
+    if metadata['Obs Type'] == 'conv':
         xlabel = conv_xlabel[metadata['Variable']]
     else:
```
CoryMartin-NOAA, line 94: Now that we have ozone, need an elif here I think

CoryMartin-NOAA: Or if metadata['Obs Type'] in ['conv', 'oz']

pyGSI/plot_diags.py
```diff
@@ -200,7 +200,7 @@ def calculate_stats(data):
     return stats
 
 
-def plot_features(dtype, stats, metadata):
+def plot_features(diag_type, stats, metadata):
```
CoryMartin-NOAA, line 203: what exactly are 'features'?

pyGSI/plot_diags.py
kevindougherty-noaa: Don't all the ozone files come as diag_<sensor>_<satellite>...? Just like the radiance files?

kevindougherty-noaa: that was weird, you cant use '< >'. What I meant was diag_sensor_satellite...

CoryMartin-NOAA: but label is 'brightness temperature', that is what I'm referring to, ozone is a mixing ratio I think (or Dobson units?)

kevindougherty-noaa: oh I see that makes sense now

pyGSI/netcdf_diags.py
kevindougherty-noaa: So the reason I did this rather than fstring is because with f string it would be a lot messier using dictionary values. Would look something like this:

ncfilename = "{metadata['Obs Type']}_{metadata['Variable']}_{metadata['ObsID']} ... etc.

pyGSI/diags.py
```diff
@@ -140,6 +140,7 @@ def __init__(self, path):
         super().__init__(path)
 
         self._read_obs()
+        self.metadata['Diag File'] = 'conventional'
```
CoryMartin-NOAA, line 143: Still sorta confused on this, should this be 'Diag File'? That suggests to me the path/filename. I would suggest 'Diag Type' but I see that is defined below...

pyGSI/diags.py
```diff
@@ -199,6 +201,12 @@ def get_data(self, diag_type, obsid=None, subtype=None, station_id=None, analysi
         OUTPUT:
             data   : requested data
         """
+        
+        self.metadata['Diag Type'] = diag_type
```
CoryMartin-NOAA, line 205: is diag_type here 'conventional' or 'temperature'? If the Former, is 'temperature' defined anywhere?

pyGSI/diags.py
```diff
@@ -488,6 +498,7 @@ def __init__(self, path):
         super().__init__(path)
 
         self._read_obs()
+        self.metadata['Diag File'] = 'radiance'
```
CoryMartin-NOAA, line 501: Same comment as above.

pyGSI/diags.py
```diff
@@ -537,6 +549,11 @@ def get_data(self, diag_type, channel=None, qcflag=None,
             data : requested data
 
         """
+        
+        self.metadata['Diag Type'] = diag_type
```
CoryMartin-NOAA, line 553: Again, what is type here, radiance, or satellite/sensor?

pyGSI/diags.py
```diff
@@ -651,6 +667,7 @@ def __init__(self, path):
         super().__init__(path)
 
         self._read_obs()
+        self.metadata['Diag File'] = 'ozone'
```
CoryMartin-NOAA, line 670: Same again.

pyGSI/diags.py
```diff
@@ -687,6 +704,9 @@ def get_data(self, diag_type, analysis_use=False, errcheck=True):
         OUTPUT:
             data   : requested data as a dictionary
         """
+        
+        self.metadata['Diag Type'] = diag_type
```
CoryMartin-NOAA, line 708: x3

pyGSI/netcdf_diags.py
```diff
@@ -109,7 +109,7 @@ def write_ncfile(stat_dict, binned_data, ncfilename, diag_type):
 
     ######################################################################
 
-    if diag_type == 'conv':
+    if obs_type == 'conv':
```
CoryMartin-NOAA, line 112: ok so I like 'obs_type' here, can we be consistent in the other file with this one?

pyGSI/netcdf_diags.py
```diff
@@ -510,51 +510,29 @@ def append_ncfile_uv(uv_stat_dict, binned_data, ncfilename):
...
-            ncfilename = '{Diag_type}_{Variable}_{ObsID[0]}_OmA.nc'.format(
-                **metadata)
-        else:
-            ncfilename = '{Diag_type}_{Satellite}_Ch{Channel[0]}_OmA.nc'.format(
+def _get_filename(metadata):
+    
+    metadata['Diag Type'] = 'omf' if metadata['Diag Type'] in ['o-f', 'omb', 'o-b'] else metadata['Diag Type']
+    metadata['Diag Type'] = 'oma' if metadata['Diag Type'] in ['o-a'] else metadata['Diag Type']        
+        
+    if metadata['Diag File'] == 'conventional':
```
CoryMartin-NOAA, line 518: I think this needs to be 'Obs Type', which I think answers my above questions. You use 'Obs Type' below in the string.

pyGSI/netcdf_diags.py
CoryMartin-NOAA: Does 'Obs Type' work in this context? Did you test this? Is it fine because you pass it a full dict?

pyGSI/netcdf_diags.py
```diff
@@ -563,39 +541,39 @@ def write_netcdf(data, binned_data, metadata):
         uv_stat_dict['Subtype'] = int(metadata['Subtype'][0])
 
         # Get the filename based on metadata
-        ncfilename = get_filename(metadata)
+        ncfilename = _get_filename(metadata)
 
-        ncpath = metadata['outDir']+'/'+ncfilename
+        ncpath = metadata['Outdir']+'/'+ncfilename
```
CoryMartin-NOAA, line 546: if the rest of the metadata keys are now spaced names, why not 'Output Dir'? or something similar?

pyGSI/netcdf_diags.py
```diff
@@ -563,39 +541,39 @@ def write_netcdf(data, binned_data, metadata):
...
     else:
 
-        stat_dict = calculate_stats(data)
+        stat_dict = _calculate_stats(data)
 
         stat_dict['Date'] = int(metadata['Date'].strftime("%Y%m%d%H"))
         stat_dict['epoch'] = metadata['Date'].timestamp()
 
-        if metadata['Diag_type'] == 'conv':
+        if metadata['Diag File'] == 'conventional':
```
CoryMartin-NOAA, line 563: Same comment/concerns as before.

pyGSI/plot_diags.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 # Created by Kevin Dougherty
 # October 2020
+# Updated January 2021
```
CoryMartin-NOAA, line 4: No need to do this because GitHub will keep the record of when things are changed.

pyGSI/plot_diags.py
```diff
@@ -250,9 +136,185 @@ def plot_features(dtype, stats, metadata):
...
+    # Observation minus Forecast
+    if metadata['Diag Type'] in ['omf', 'o-f']:
+        xlabel = 'Observation - Forecast'
+    
+    # Observation minus Analysis
+    elif metadata['Diag Type'] == 'oma':
+        xlabel = 'Observation - Analysis'
+    
+    # Conventional Data
+    elif metadata['Diag File'] == 'conventional':
```
CoryMartin-NOAA, line 176: Flagging for awareness, think this should be fixed or at least discussed.

pyGSI/plot_diags.py
```diff
@@ -250,9 +136,185 @@ def plot_features(dtype, stats, metadata):
...
+    else:
+        xlabel = None
+        
+    
+    return xlabel
+
+def _get_stats_labels(metadata,stats):
+    
+    if metadata['Diag File'] == 'conventional' and metadata['Variable'] == 'q':
+        roundnum = 6
```
CoryMartin-NOAA, line 201: is this the number of digits to round to? Is there a better name than 'roundnum'? I guess it's probably okay

pyGSI/plot_diags.py
```diff
@@ -250,9 +136,185 @@ def plot_features(dtype, stats, metadata):
...
+    return t
+    
+def _get_title(metadata):
+    
+    # Handles analysis use data
+    anl_use = True if 'Anl Use' in metadata else False  
+    
+    # Left Title label
+    if anl_use:
+        if metadata['Diag File'] == 'conventional':
```
CoryMartin-NOAA, line 223: here's another spot.

pyGSI/plot_diags.py
```diff
@@ -250,9 +136,185 @@ def plot_features(dtype, stats, metadata):
...
+
+def _get_save_file(metadata):
+    
+    # Save file label
+    if metadata['Diag File'] == 'conventional':
+        save_file = '{Date:%Y%m%d%H}_{Obs Type}_{Variable}_{Diag Type}_'.format(
+            **metadata) + '%s' % '_'.join(str(x) for x in metadata['ObsID'])
+    
+    elif metadata['Diag File'] == 'ozone':
+        save_file = '{Date:%Y%m%d%H}_{Obs Type}_{Satellite}_{Diag Type}_{Diag File}'.format(
```
CoryMartin-NOAA, line 261: I don't think we need the filename for the figure to end in ozone, it will be implied by ompslp...blah blah whatever for obs type and satellite, which should be 'sensor' and 'satellite'

scripts/create_conv_netcdf.py
```diff
@@ -33,20 +33,20 @@ def first_occurrence(worklist):
 
 
 def create_netcdf(conv_config):
-
+    
     diagfile = conv_config['conventional input']['path'][0]
```
CoryMartin-NOAA, line 37: see this proves my point about the confusion of the name of the key in the dictionary above

CoryMartin-NOAA (changes requested):
My main comment is I think there's still some confusion about 'Diag File', which to me suggests the path to the netCDF file, and I think it should be more like 'Obs Type', for conventional, radiance, or ozone. Then places where Obs Type are used (ozone only?) you need sensor and satellite together, and for conventional you use variable, ObsID and subID (or whatever they're called) to indicate radiosondes, surface, etc.

pyGSI/diags.py
kevindougherty-noaa: So my main motivation for this is because I needed to separate by diagnostic file type conventional, ozone, or radiance. I cannot get this information straight from the file other than conventional where 'Obs Type' from the file name would be 'conv'. However, this gets complicated with radiance and ozone because 'Obs Type' from the filename is the sensor name. I would then have to distinguish all the different sensors and satellites between radiance and ozone.

pyGSI/diags.py
kevindougherty-noaa: 'Diag Type' is omf, observation, hofx, etc. 'Variable' would be like temperature, q, uv etc.

pyGSI/diags.py
CoryMartin-NOAA: I get the reason for this, but Diag File isn't the right label/name is what I'm saying. What about self.metadata['conventional'] = True? and self.metadata['radiance'] = False and self.metadata['ozone'] = False?

kevindougherty-noaa: Yes I can do that, but will need to make sure to include all 3 in every instance otherwise will need to add a similar line like when Anl Use is added: `anl_use = True if 'Anl Use' in metadata else False`

CoryMartin-NOAA: yeah that's probably more work than is needed. What about just 'Diag File Type'? Just so it's clearer

CoryMartin-NOAA: or just 'File Type'

kevindougherty-noaa: So File Type is already taken because that is either 'ges' or 'anl' from the filename. So 'Diag File Type' works with me. I understand the confusion, there's a lot of different variable names that are in play here.

CoryMartin-NOAA:
@kevindougherty-noaa do we think every comment has been addressed and it's ready to merge? I think so, but I want to double check with you.

kevindougherty-noaa:
@CoryMartin-NOAA Yes, the comments have been addressed and I believe it is ready to be merged. Let me change the shell scripts to be up to date with new changes and it will be ready to be added.

CoryMartin-NOAA (approved):

Issue #30, opened 2021-01-19
Add support to plot RMS, bias profiles, etc. from gsistat files


CoryMartin-NOAA:

CoryMartin-NOAA:
I'll take the lead but consult @CatherineThomas-NOAA to verify the results are the same as the original example script.

CoryMartin-NOAA:
![Screen Shot 2021-01-20 at 1 39 52 PM](https://user-images.githubusercontent.com/6354668/105219925-058d5c80-5b25-11eb-8cfa-6ac42c1335ac.png)

Sample using @CatherineThomas-NOAA 's control and experiment from June 2020 dropout case, will issue PR soon.

PR #31, opened 2021-01-20
Add script to plot experiment RMSE, bias, counts from GSI stat file

CoryMartin-NOAA:
@aerorahul had a two-step workflow to parse and plot GSI stat files, and later introduced a GSI stat Python class to simplify this process. This PR adds a script that uses the PyGSI GSI stat class to read gsistat files, and then plots bias, RMSE, and counts for a control and N experiments (currently just O-F profiles of radiosondes, but should be easily adaptable in the future)

NOTE: this will likely fail if you run 'out of the box', please make sure you add the path of this cloned repo (/path/to/clone/PyGSI) to your $PYTHONPATH, so when you run the script, the python interpreter can find the files needed by 'import pygsi'

This closes issue #30

### DIFF
--- a/scripts/plot_gsi_stat_exp.py
+++ b/scripts/plot_gsi_stat_exp.py
@@ -15,65 +15,74 @@ obtypes = [120, 220]
 obvars = ['t', 'uv', 'q']
 levels = [1000, 900, 800, 600, 400, 300, 250, 200, 150, 100, 50, 0]
 
+
 def gen_figure(datadict, datatypestr, labels, sdate, edate, save, plotdir):
     # Line/marker colors for experiments ('k' is the first)
-    mc = ['k', 'r', 'g', 'b', 'm','c','y']
+    mc = ['k', 'r', 'g', 'b', 'm', 'c', 'y']
 
     # set figure params one time only.
     rcParams['figure.subplot.left'] = 0.1
-    rcParams['figure.subplot.top']  = 0.85
-    rcParams['legend.fontsize']     = 12
-    rcParams['axes.grid']           = True
+    rcParams['figure.subplot.top'] = 0.85
+    rcParams['legend.fontsize'] = 12
+    rcParams['axes.grid'] = True
 
-    fig1 = plt.figure(figsize=(10,8))
+    fig1 = plt.figure(figsize=(10, 8))
     plt.subplots_adjust(hspace=0.3)
-    gs = gspec.GridSpec(1,3)
+    gs = gspec.GridSpec(1, 3)
     stattype = 'mean' if datatypestr.lower() in ['bias', 'rmse'] else 'sum'
 
     for v, var in enumerate(obvars):
         xmin = 999
         xmax = 0
         ax = plt.subplot(gs[v])
-        for e,expid in enumerate(labels):
+        for e, expid in enumerate(labels):
             profile = datadict[expid][stattype][var][:-1]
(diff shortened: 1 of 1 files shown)


scripts/plot_gsi_stat_exp.py
```diff
@@ -0,0 +1,160 @@
...
+    for exp, gsistats in zip(labels, args.gsistats):
+        rmses[exp] = {}; counts[exp] = {}; biases[exp] = {}
+        for gsistat, cycle in zip(statfiles, cycles):
+            rmses[exp][cycle] = {}; counts[exp][cycle] = {}; biases[exp][cycle] = {}
+            inputfile = os.path.join(gsistats, gsistat)
+            try:
+                gdas = GSIstat(inputfile, cycle)
+            except FileNotFoundError:
+                print('Error: One or more expected gsistat files missing, check dates and paths!')
+                sys.exit(1)
```
aerorahul, line 131: use:
```
raise FileNotFoundError(f'Unable to find {inputfile} for cycle {cycle}')
```

scripts/plot_gsi_stat_exp.py
```diff
@@ -0,0 +1,160 @@
...
+        mydate = mydate + datetime.timedelta(hours=6)
+    return statfiles, cycles
+
+if __name__ == '__main__':
+    # get command line arguments
+    parser = argparse.ArgumentParser(description=('Plots a comparison of GSI ',
+                                                  'statistics for one or more ',
+                                                  'experiments compared to a ',
+                                                  'reference control run.'))
+    parser.add_argument('-g','--gsistats',help='list of directories containing GSI stat files',nargs='+',required=True)
```
aerorahul, line 99: does this need to be `-g`? It is directories containing GSI stat files; `-d` not good enough?

scripts/plot_gsi_stat_exp.py
```diff
@@ -0,0 +1,160 @@
+#!/usr/bin/env python
```
aerorahul, line 1: Pass this through pep8.

aerorahul (commented):
looks good, but I did not test.
some minor comments.

scripts/plot_gsi_stat_exp.py
CoryMartin-NOAA: Done.

scripts/plot_gsi_stat_exp.py
CoryMartin-NOAA: Yeah -d works for me, fixed.

scripts/plot_gsi_stat_exp.py
CoryMartin-NOAA: Ok added this.

aerorahul (approved):
looks good, but I didn't test this.

CoryMartin-NOAA:
I tested and put an output figure in issue #30 but I would like @CatherineThomas-NOAA to test/review before we merge in.

CatherineThomas-NOAA (approved):
Changes look good to me.  I tested it on Hera, compared to a plot from original script, and it works as expected.  See the attached pdf for a comparison.

[Gsistat utility.pdf](https://github.com/NOAA-EMC/PyGSI/files/5850860/Gsistat.utility.pdf)

CoryMartin-NOAA:
Thanks @CatherineThomas-NOAA ! I'll merge this in.

PR #32, opened 2021-01-29
feature/mp_ozone

kevindougherty-noaa:
added mp_ozone_diags.py, edited plot_diags.py to change plot title and save file, and added analysis_use to ozinfo2yaml.py


pyGSI/plot_diags.py
```diff
@@ -254,15 +259,17 @@ def _get_save_file(metadata):
     
     # Save file label
     if metadata['Diag File Type'] == 'conventional':
-        save_file = '{Date:%Y%m%d%H}_{Obs Type}_{Variable}_{Diag Type}_'.format(
+        save_file = '{Date:%Y%m%d%H}_{Obs Type}_{Variable}_{Diag Type}'.format(
```
CoryMartin-NOAA, line 262: are these supposed to have suffixes? If so, I don't see where it is added/what the final form of the save_file will be.

CoryMartin-NOAA: Or is this just for the label on the figure? If this, is there a better name than 'save_file'?

scripts/mp_ozone_diags.py
```diff
@@ -0,0 +1,77 @@
...
+import yaml
+from multiprocessing import Pool
+import sys
+from pyGSI.diags import Ozone
+from pyGSI.plot_diags import plot_spatial, plot_histogram
+from datetime import datetime
+
+start_time = datetime.now()
+
+def plotting(ozone_config):
```
CoryMartin-NOAA, line 12: How is this file different from the other file? mp_plot_ozone_diags.py  Is this a mistake/error in filename change?

kevindougherty-noaa: yes, I named it wrong at first adn forgot to delete the other one.

pyGSI/plot_diags.py
kevindougherty-noaa: This is the saved file name. The suffixes come from the metadata dictionary.

CoryMartin-NOAA: Yeah I see now, got it!

CoryMartin-NOAA (approved):

PR #41, opened 2021-03-04
Feature/diags pandas

kevindougherty-noaa:
I am currently updating diags.py to utilize pandas. I am working one class at a time to save reviewers from going over hundreds of lines of code. The first class added is Conventional, and will be followed by Radiance, and then Ozone. Pandas allows for much better readability for future users to better understand how the data is being indexed. Creating an indexed dataframe that explicitly calls columns for the user to know exactly what is being filtered. 

I also cleaned up get_data( ) and added _pressure_indexing( ) for more simplicity.

### DIFF
--- a/pyGSI/diags.py
+++ b/pyGSI/diags.py
@@ -1,4 +1,6 @@
 import os
+import xarray as xr
+import pandas as pd
 import numpy as np
 from netCDF4 import Dataset
 from datetime import datetime
@@ -13,7 +15,7 @@ class GSIdiag:
         INPUT:
             path : path to GSI diagnostic object
         """
-        
+
         self.path = path
         self.filename = os.path.splitext(Path(self.path).stem)[0]
         self.obs_type = self.filename.split('_')[1]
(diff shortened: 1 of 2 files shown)


pyGSI/diags.py
```diff
@@ -59,67 +61,69 @@ def query_diag_type(self, diag_type, idx):
...
+                    indexed_df['v_Obs_Minus_Forecast_adjusted']
 
-                return u, v
+                return u.to_numpy(), v.to_numpy()
 
             else:
-                hx = self.o - self.omf
-                data = hx[idx]
+                data = indexed_df['Observation'].to_numpy(
+                ) - indexed_df['Obs_Minus_Forecast_adjusted'].to_numpy()
```
CoryMartin-NOAA, line 106: Does this pass pep8 compliance? If it does, can you move the ) to be inline with the indexed_ on the above line? I feel like it's hard to read with the line indented the same as the data = part (does that make sense? add 7 spaces to front of  this line)

CoryMartin-NOAA (changes requested):
There are changes to multiple classes here, is that correct? Also am I correct in that now there is no longer self.lats (for example)? How do the users get the lat, lon, etc. values?

kevindougherty-noaa:
When committing new changes, I made a mistake of deleting the classes that already existed. Closing PR to figure out how to revert back to original.

PR #42, opened 2021-03-04
Feature/pandas diags

kevindougherty-noaa:
Added pandas to diags.py classes to allow for much better readability and better understanding on how indexing works for future users. I will be adding changes to each class in this order: Conventional, Radiance, Ozone.

### DIFF
--- a/pyGSI/diags.py
+++ b/pyGSI/diags.py
@@ -6,7 +6,6 @@ from netCDF4 import Dataset
 from datetime import datetime
 from pathlib import Path
 
-
 class GSIdiag:
 
     def __init__(self, path):
@@ -15,7 +14,7 @@ class GSIdiag:
         INPUT:
             path : path to GSI diagnostic object
         """
-
+        
         self.path = path
         self.filename = os.path.splitext(Path(self.path).stem)[0]
         self.obs_type = self.filename.split('_')[1]
@@ -36,100 +35,38 @@ class GSIdiag:
     def __len__(self):
         return len(self.lats)
 
-    def query_diag_type(self, diag_type, indexed_df):
+    def _query_diag_type(self, diag_type, df, bias_correction=False):
         """
         Query the data type being requested and returns
         the appropriate indexed data
         """
-
+        
         diag_type = diag_type.lower()
-
-        # Determines the diagnostic type
-        omf = True if diag_type in ['omf', 'o-f', 'omb', 'o-b'] else False
-        oma = True if diag_type in ['oma', 'o-a'] else False
-        obs = True if diag_type in ['observation'] else False
-        hofx = True if diag_type in ['hofx'] else False
-        water_fraction = True if diag_type in ['water_fraction'] else False
-        land_fraction = True if diag_type in ['land_fraction'] else False
-        cloud_fraction = True if diag_type in ['cloud_fraction'] else False
-        snow_fraction = True if diag_type in ['snow_fraction'] else False
-        ice_fraction = True if diag_type in ['ice_fraction'] else False
-
-        # Determines the file type
-        is_ges = True if self.ftype == 'ges' else False
-        is_anl = True if self.ftype == 'anl' else False
-
-        if omf and is_ges:
+        
+        diag_type = 'omf' if diag_type in ['omf', 'o-f', 'omb', 'o-b', 'o-a', 'oma'] else diag_type
+        bias = 'adjusted' if bias_correction else 'unadjusted'
+        
+        try:
             if self.variable == 'uv':
-                u = indexed_df['u_Obs_Minus_Forecast_adjusted'].to_numpy()
-                v = indexed_df['v_Obs_Minus_Forecast_adjusted'].to_numpy()
-
+                try:
+                    u = df[f'u_{diag_type}_{bias}']
+                    v = df[f'v_{diag_type}_{bias}']
+                except:
+                    u = df[f'u_{diag_type}']
+                    v = df[f'v_{diag_type}']
+                    
                 return u, v
-
             else:
-                data = indexed_df['Obs_Minus_Forecast_adjusted'].to_numpy()
+                try:
+                    data = df[f'{diag_type}_{bias}']
+                except:
+                    data = df[f'{diag_type}']
+                
                 return data
-
-        if oma and is_anl:
-            if self.variable == 'uv':
-                u = indexed_df['u_Obs_Minus_Forecast_adjusted'].to_numpy()
-                v = indexed_df['v_Obs_Minus_Forecast_adjusted'].to_numpy()
-
-                return u, v
-
-            else:
-                data = indexed_df['Obs_Minus_Forecast_adjusted'].to_numpy()
-                return data
-
-        if obs:
-            if self.variable == 'uv':
-                u = indexed_df['u_Observation'].to_numpy()
-                v = indexed_df['v_Observation'].to_numpy()
-
-                return u, v
-
-            else:
-                data = indexed_df['Observation'].to_numpy()
-                return data
-
-        if hofx:
-            if self.variable == 'uv':
-                u = indexed_df['u_Observation'] - \
-                    indexed_df['u_Obs_Minus_Forecast_adjusted']
-                v = indexed_df['v_Observation'] - \
-                    indexed_df['v_Obs_Minus_Forecast_adjusted']
-
-                return u.to_numpy(), v.to_numpy()
-
-            else:
-                data = indexed_df['Observation'].to_numpy() - \
-                    indexed_df['Obs_Minus_Forecast_adjusted'].to_numpy()
-                return data
-
-        if water_fraction:
-            data = indexed_df['Water_Fraction'].to_numpy()
-            return data
-
-        if land_fraction:
-            data = indexed_df['Land_Fraction'].to_numpy()
-            return data
-
-        if ice_fraction:
-            data = indexed_df['Ice_Fraction'].to_numpy()
-            return data
-
-        if snow_fraction:
-            data = indexed_df['Snow_Fraction'].to_numpy()
-            return data
-
-        if cloud_fraction:
-            data = indexed_df['Cloud_Frac'].to_numpy()
-            return data
-
-        else:
+        except:
             raise Exception(f'Unrecognizable diag type input: {diag_type}')
             return None
-
+            
 
 class Conventional(GSIdiag):
 
@@ -148,7 +85,7 @@ class Conventional(GSIdiag):
 
     def __str__(self):
         return "Conventional GSI diagnostic object"
-
+    
     def _read_obs(self):
         """
         Reads the data from the conventional diagnostic file during initialization.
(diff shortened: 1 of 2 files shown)


pyGSI/diags.py
```diff
@@ -316,173 +237,180 @@ def get_data(self, diag_type, obsid=None, subtype=None, station_id=None, analysi
...
 
                     return u, v
 
                 else:
-                    data = self.query_diag_type(diag_type, idx)
+                    data = self.query_diag_type(diag_type, indexed_df)
 
                     return data
 
-    def _get_idx_conv(self, obsid=None, subtype=None, station_id=None, analysis_use=False):
```
CoryMartin-NOAA, line 332: Why was analysis_use removed here? I don't necessarily see it as a problem but want to know your reasoning.

kevindougherty-noaa: So my reasoning is that it makes it easier and cleaner to do pressure indexing. When trying to create the binning with assimilated and monitored data, it was difficult to assign which was which when only given one dataframe in _pressure_indexing(). Makes it much easier to do the pressure indexing across just regular data, lat/lon, and pressure data.

CoryMartin-NOAA: What if you don't want to do pressure indexing though? Then can you still do analysis_use? Or not?

aerorahul (commented):
I have some questions on the design of this now that dataframes are used.
A dataframe object contains all the info of a diag file and so it is very easy to find the data by just keying in the key.
We probably don't need a lot of the if-else to return data.

if you want a method to get the data, it can be as easy as:

```
def get_data_from_dataframe(df, key):
    try:
           data = df['key']
    except:
            raise KeyValueError('dataframe does not contain {key}')
      return data
```

To then use this:
```
data = get_data_from_dataframe(df, key)
```

Thoughts.

pyGSI/diags.py
kevindougherty-noaa: Oh I see. Analysis_use is still addressed in lines 210-212 in get_data().
```
if analysis_use:
    assimilated_df = indexed_df.loc[df['Analysis_Use_Flag'] == 1]
    monitored_df = indexed_df.loc[df['Analysis_Use_Flag'] == -1]
```

CoryMartin-NOAA: ok got it.

CoryMartin-NOAA:
@aerorahul makes a good point, but I would think we still want some level of abstraction between the raw variable names and what the user is looking for, like 'u_Obs_Minus_Forecast_adjusted' is annoying to type when you can just have the user as for OmF, or whatever.

pyGSI/diags.py
```diff
@@ -59,67 +61,69 @@ def query_diag_type(self, diag_type, idx):
 
         if omf and is_ges:
             if self.variable == 'uv':
-                u = self.u_omf[idx]
-                v = self.v_omf[idx]
+                u = indexed_df['u_Obs_Minus_Forecast_adjusted'].to_numpy()
+                v = indexed_df['v_Obs_Minus_Forecast_adjusted'].to_numpy()
```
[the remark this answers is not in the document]
kevindougherty-noaa: `.to_numpy() ` returns the data as an array, staying consistent with the original code and not creating issues with plotting/spatial bin scripts

aerorahul (commented):
Also, the point of using pandas is to minimize using variables.  One should be able to access the variables from the dataframe object directly.
When one needs subset of variables, create a new dataframe so that one maintains the structure and that allows one to use indices from one dataframe onto another.

pyGSI/diags.py
```diff
@@ -59,67 +61,69 @@ def query_diag_type(self, diag_type, idx):
...
 
         if snow_fraction:
-            data = self.snow_frac[idx]
+            data = indexed_df['Snow_Fraction'].to_numpy()
             return data
 
         if cloud_fraction:
-            data = self.cloud_frac[idx]
+            data = indexed_df['Cloud_Frac'].to_numpy()
             return data
```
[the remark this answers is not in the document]
CoryMartin-NOAA: Yeah perhaps a dictionary of key, value pairs that can take the cloud_fraction string from before and 'translate' it to the 'Cloud_Frac' name used in the GSI diag?

aerorahul:
> @aerorahul makes a good point, but I would think we still want some level of abstraction between the raw variable names and what the user is looking for, like 'u_Obs_Minus_Forecast_adjusted' is annoying to type when you can just have the user as for OmF, or whatever.

if you do that, you made a copy of the data.
Not to mention, OmF is ambiguous in this case as it does not say adjusted, which is what the user extracted.
May be the dataframe itself can be renamed from `u_Obs_Minus_Forecast_adjusted` to something short.  
ps. what is `u_` and why?

kevindougherty-noaa:
I agree with both of you. Rahul, I think that is great idea to git rid of all the if/else statements now that we have the dataframe. Cory, I still think it is also important to keep the variables simple i.e. 'omf', 'hofx' etc.

CoryMartin-NOAA:
if we can rename them easily enough in the data frame, then I support that approach.

@aerorahul u is the eastward wind component ;-)

pyGSI/diags.py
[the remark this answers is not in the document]
aerorahul: you should not need that at all.
if you need this at the time of plotting etc call this method where needed and eventually remove it.
I don't recall having to use it anywhere when I was doing binning/plotting/averaging etc.

pyGSI/diags.py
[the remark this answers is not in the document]
aerorahul: If I recall, these are keys coming directly from xarray reading of the netcdf file.
I don't know if we want to create mappings of what is in the file to what is "convenient".  It just creates confusion.
WYSIWYG. Agreed those keys are a bit long, but we put them there.  We could have used `long_name` in the netcdf to describe the variable attribute.

aerorahul:
> if we can rename them easily enough in the data frame, then I support that approach.
> 
> @aerorahul u is the eastward wind component ;-)

I thought it was some unicode character being added.  I've been away from this far too long.

pyGSI/diags.py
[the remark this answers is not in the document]
CoryMartin-NOAA: I'm 90% confident there is no long_name attribute for these files...

CoryMartin-NOAA:
Every other observation is just observation_... but uv has u_observation and v_observation, so they have to be handled separately.

aerorahul:
Another thing that we can do is break the original dataframe into multiple ones.
e.g.
a dataframe for OmF, OmA OmF_adjusted etc.
They all will have the same indices and columns.

so when you call `OmF['u']` it is the `u_Observation_minus Forecast` and so on.

where can I find a netcdf diag file and load it?  Can someone point to one on Orion.

aerorahul:
> @aerorahul makes a good point, but I would think we still want some level of abstraction between the raw variable names and what the user is looking for, like 'u_Obs_Minus_Forecast_adjusted' is annoying to type when you can just have the user as for OmF, or whatever.

how else would you do:
```
u = get_data_from_dataframe(df, 'u_Obs_Minus_Forecast_adjusted')
```
you have to type the whole thing somewhere, unless you renamed `u_Obs_Minus_Forecast_adjusted` to `u_OmF_adj` right after populating the dataframe.

CoryMartin-NOAA:
On Orion, /work/noaa/da/Cory.R.Martin/noscrub/GFS_sample/gdas.20200328/12/diag_conv_uv_ges.2020032812.nc4

Well, before, to the user, this was transparent, they just ask for OmF, the full string is only typed once, when the PyGSI API is created.

pyGSI/diags.py
```diff
@@ -34,98 +35,38 @@ def __init__(self, path):
...
-
-                return u, v
-
-            else:
-                data = self.o[idx]
-                return data
-
-        if hofx:
+        
+        diag_type = 'omf' if diag_type in ['omf', 'o-f', 'omb', 'o-b', 'o-a', 'oma'] else diag_type
```
aerorahul, line 46: This is not an appropriate use of the ternary conditional operator where the else is the same as the LHS.

pyGSI/diags.py
```diff
@@ -34,98 +35,38 @@ def __init__(self, path):
     def __len__(self):
         return len(self.lats)
 
-    def query_diag_type(self, diag_type, idx):
+    def _query_diag_type(self, diag_type, df, bias_correction=False):
```
aerorahul, line 38: `_query_diag_type` indicates it is a private method and yet it is explicitly called from a sub-class.

aerorahul: It is also returning the data not just setting diag_type.

pyGSI/diags.py
```diff
@@ -144,345 +85,195 @@ def __init__(self, path):
...
+    
     def _read_obs(self):
         """
         Reads the data from the conventional diagnostic file during initialization.
+        Stores the data as a pandas dataframe.
         """
+        ds = xr.open_dataset(self.path)
+        df = ds.to_dataframe()
+        
+        # Covert byte strings to strings
```
aerorahul, line 97: Not covert

pyGSI/diags.py
```diff
@@ -144,345 +85,195 @@ def __init__(self, path):
...
             idx    : indices of the requested data in the file
         """
-
+        
         if not obsid or obsid == [None]:
             obsid = None
         if not subtype or subtype == [None]:
             subtype = None
         if not station_id or station_id == [None]:
             station_id = None
```
aerorahul, line 219: you don't need it at all.
The defaults are provided at the function interface.

aerorahul (changes requested):
some comments.
I don't quite understand what `indexed_df` is?  It is already indexed when read in in `_read_obs` function.

pyGSI/diags.py
kevindougherty-noaa: Thank you for catching that. Do you suggest renaming it `return_diag_type_data()` or something along those lines to be more descriptive?

pyGSI/diags.py
kevindougherty-noaa: I see now. Pandas handles better if for example obsid = [None] which can happen if using a YAML compared to the previous version written.

pyGSI/diags.py
kevindougherty-noaa: > some comments.
> I don't quite understand what `indexed_df` is? It is already indexed when read in in `_read_obs` function.

I guess my wording is not great. The df is made into a multiindexed df in the `read_obs` method by the columns in the list. The indexed_df is the df where all of the inputs by the user (obsid, subtype, station_id) are valid. I should probably change the method name because I am not returning the index of where these are valid like the original script was, I am now returning the already valid dataframe (`indexed_df`). Please let me know if that makes sense.

pyGSI/diags.py
```diff
@@ -144,345 +85,195 @@ def __init__(self, path):
...
         self.metadata['Subtype'] = subtype
         self.metadata['Station ID'] = station_id
         self.metadata['Anl Use'] = analysis_use
+        
+        if analysis_use:
+            assimilated_df, monitored_df = self._get_idx_conv(obsid, subtype, station_id, analysis_use)
+            
+            if self.variable == 'uv':
+                u_assimilated, v_assimilated = self._query_diag_type(
+                    diag_type, indexed_df, bias_correction)
```
aerorahul, line 165: what is `indexed_df` and where is it coming from in this method?

CoryMartin-NOAA:
@kevindougherty-noaa there are conflicts, need to merge in 'main' to your branch after the changes Xiaoyan added.

pyGSI/diags.py
kevindougherty-noaa: That is a copy and paste mistake. Should be assimilated_df and monitored_df respectively which is called just above. Changing now.

pyGSI/diags.py
```diff
@@ -144,345 +85,195 @@ def __init__(self, path):
...
-                    pres_idx = np.where((self.press > pressure_list[i]) & (
-                        self.press < pressure_list[i+1]))
-                    valid_idx = np.isin(idx[0], pres_idx[0])
-                    pidx = np.where(valid_idx)
-
-                    pressure_lats[pressure] = self.lats[pidx]
-                    pressure_lons[pressure] = self.lons[pidx]
-
-                return pressure_lats, pressure_lons
+            lats = {'assimilated': assimilated_pressure_df['latitude'],
```
aerorahul, line 249: what is `assimilated_pressure_df` and `monitored_pressure_df`

aerorahul:
@kevindougherty-noaa 
There are several issues in this PR that need a careful check before it is really reviewable.

pyGSI/diags.py
CoryMartin-NOAA: Why does it need to include 'diag type'? It's part of the class, so can't it just be 'get_data' or something like that?

pyGSI/diags.py
CoryMartin-NOAA: He's saying you have diag_type here too many times (I think).
It needs to be:
diag_type = 'omf' if DiagType in [...] else diag_type

I think.

Because it will overwrite diag_type's previous value with what is in the list.

pyGSI/diags.py
CoryMartin-NOAA: Well those byte strings are sneaky

pyGSI/diags.py
CoryMartin-NOAA: So I see get_data is below. Is this to get the entire array of input data and the latter one is the requested subset? Perhaps this is better as combined all together? Pass diag_type to the get_data routine?

pyGSI/diags.py
CoryMartin-NOAA: Yeah I have the same question, where are they defined? I don't see them as input to the function.

CoryMartin-NOAA (commented):
Some initial comments from me as well.

pyGSI/diags.py
kevindougherty-noaa: The reason we do not combine the two is because the current `query_diag_type()` applies to all classes eventually, but right now I am only focusing on making changes to `Conventional()`. The name of the method needs to change because we used to literally query through all different diag types. Now with the pandas df, we can just use the diag_type name to extract the correct subsetted column from the dataframe. I will rename the method to reflect what it does now more accurately.

pyGSI/diags.py
```diff
@@ -34,98 +35,40 @@ def __init__(self, path):
...
-        if snow_fraction:
-            data = self.snow_frac[idx]
-            return data
-
-        if cloud_fraction:
-            data = self.cloud_frac[idx]
-            return data
-
-        else:
+        except:
```
CoryMartin-NOAA, line 68: Why is this wrapped in a try, except and then an if? why not just raise exception in the else? Am I missing something?

pyGSI/diags.py
```diff
@@ -144,345 +87,197 @@ def __init__(self, path):
...
-
-            idx = np.where(valid_idx)
-
-            return idx
-
+        
+        df = self.data_df
+        
+        if obsids is not None:
+            indx = df.index.get_level_values('Observation_Type') == ''
```
CoryMartin-NOAA, line 215: Is this first one just so the loop doesn't error on the first iteration?

CoryMartin-NOAA (commented):
A few comments/questions.

pyGSI/diags.py
```diff
@@ -34,98 +35,40 @@ def __init__(self, path):
...
+        
+        try:
             if self.variable == 'uv':
-                u = self.u_o[idx] - self.u_omf[idx]
-                v = self.v_o[idx] - self.v_omf[idx]
-
+                try:
+                    u = df[f'u_{diag_type}_{bias}']
+                    v = df[f'v_{diag_type}_{bias}']
+                except:
```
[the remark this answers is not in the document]
kevindougherty-noaa: It's more or less just tries to grab the column if it is omf and including the bias variable in the column, otherwise it does not include the bias variable. I understand what you are saying though and can figure out what type of error occurs if column does not exist. This helps to eliminate things like if omf, do something else other variable, do something.

pyGSI/diags.py
kevindougherty-noaa: It is wrapped like this in the case the user inputs a diag_type that does not make sense. i.e. diag_type = 'blah'

pyGSI/diags.py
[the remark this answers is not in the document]
CoryMartin-NOAA: Yeah, my point was what you have is fine, but what if there is some other sort of error that we want to cause it to fail. 'except' should generally only be explicit (and can be multiple types, I forget the exact syntax though of how to do that)

pyGSI/diags.py
CoryMartin-NOAA: Ok I see, so this is an excellent illustration of my earlier point, then this should be 'except KeyError' or something similar.

pyGSI/diags.py
kevindougherty-noaa: This was taken from Rahul's code that he shared. It returns an array of all booleans that are False, then loops through every input of the variable and finds where the variable is true and replaces the index with that index. That way, if there is no cases of that variable, the index remains all false and is applied to the dataframe. That way, if one of the inputs does not exist, the dataframe will return as empty.

pyGSI/diags.py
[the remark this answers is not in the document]
kevindougherty-noaa: Got it. tested it out and it was KeyError so I will include that in next commit.

pyGSI/diags.py
```diff
@@ -34,98 +35,40 @@ def __init__(self, path):
     def __len__(self):
         return len(self.lats)
 
-    def query_diag_type(self, diag_type, idx):
+    def get_diag_type_data(self, diag_type, df, bias_correction=False):
```
aerorahul, line 38: why do we need df? Isn't `data_df` and attribute of the object?

pyGSI/diags.py
aerorahul: if a user inputs a `blah` diag_type, it should never get here.

aerorahul:
I am thinking if we can have a way to hide the 'if-else' with regards to uv and everything else

pyGSI/diags.py
aerorahul: @kevindougherty-noaa is correct in explaining why the first line "initializes" the search locations to empty.  In the end, if there are no matches, it will return nothing.  That is the point of 'selecting' data based on search criteria.

pyGSI/diags.py
kevindougherty-noaa: yes, but when selecting data, if analysis_use=True, `_select_conv()` returns assimilated_df and monitored_df. So unless we address analysis_use in `get_diag_tyep_data()`, we need df

CoryMartin-NOAA:
This is really old, closing.

PR #47, opened 2021-03-17
Bugfix/plot diags

kevindougherty-noaa:
Missing ':' on line 27.


CoryMartin-NOAA (approved):
Whoops, this is my bad. Good catch/fix.

PR #48, opened 2021-03-26
Feature/stnid fix

kevindougherty-noaa:
fixed stnid indexing because old version was deprecated.

### DIFF
--- a/pyGSI/diags.py
+++ b/pyGSI/diags.py
@@ -361,8 +361,8 @@ class Conventional(GSIdiag):
                 valid_idx = np.logical_and(valid_idx, valid_subtype_idx)
             if station_id != None:
                 stnididx = self.stnid
-                stnididx = np.array(
-                    [''.join(i.tostring().decode().split()) for i in stnididx])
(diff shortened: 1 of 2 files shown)


pyGSI/diags.py
```diff
@@ -361,8 +361,8 @@ def _get_idx_conv(self, obsid=None, subtype=None, station_id=None, analysis_use=
                 valid_idx = np.logical_and(valid_idx, valid_subtype_idx)
             if station_id != None:
                 stnididx = self.stnid
-                stnididx = np.array(
-                    [''.join(i.tostring().decode().split()) for i in stnididx])
+                stnididx = [i.tobytes(fill_value='/////', order='C') for i in stnididx]
```
CoryMartin-NOAA, line 364: Any reason why ///// was chosen? Will this cause any issues? Probably not.

CoryMartin-NOAA (approved):

pyGSI/diags.py
kevindougherty-noaa: No, I was following how the old version filled the empty values.

Issue #56, opened 2021-07-20
Allow data binning by height.


EdwardSafford-NOAA:
PyGSI supports binning data using an input list of pressure levels.  For some data sources, specifically gnss, height (altitude) is a more meaningful unit for binning.  Extend the existing pyGSI components by adding a bin type to enable support for both pressure and height.

CoryMartin-NOAA:
Closed by PR #57

PR #57, opened 2021-07-22
PyGSI issue #56, add binning by height to class Conventional.

EdwardSafford-NOAA:
Modified class conventional to bin data by height instead of pressure.   The default (assuming binning is used) is pressure.  I'm working on adding vertical plots for gps data to the package and height (altitude) makes much more sense for gps data than does pressure.  This extension is a prerequisite for those plots.

I've run scripts/batch_conv_diags.sh both with the original code and with my changes and the results are identical.

### DIFF
--- a/pyGSI/diags.py
+++ b/pyGSI/diags.py
@@ -4,6 +4,7 @@ from netCDF4 import Dataset
 from datetime import datetime
 from pathlib import Path
 
+_VALID_LVL_TYPES = ["pressure", "height"]
 
 class GSIdiag:
 
@@ -142,6 +143,7 @@ class Conventional(GSIdiag):
         self._read_obs()
         self.metadata['Diag File Type'] = 'conventional'
 
+
     def __str__(self):
         return "Conventional GSI diagnostic object"
 
@@ -156,9 +158,11 @@ class Conventional(GSIdiag):
             self.lons = f.variables['Longitude'][:]
             self.lats = f.variables['Latitude'][:]
             self.press = f.variables['Pressure'][:]
+            self.height = f.variables['Height'][:]
             self.time = f.variables['Time'][:]
             self.anl_use = f.variables['Analysis_Use_Flag'][:]
             self.stnid = f.variables['Station_ID'][:]
+
             try:
                 self.stnelev = f.variables['Station_Elevation'][:]
             except:
@@ -176,7 +180,8 @@ class Conventional(GSIdiag):
                 self.omf = f.variables['Obs_Minus_Forecast_adjusted'][:]
 
 
-    def get_data(self, diag_type, obsid=None, subtype=None, station_id=None, analysis_use=False, plvls=None):
+
+    def get_data(self, diag_type, obsid=None, subtype=None, station_id=None, analysis_use=False, lvls=None, lvl_type='pressure'):
         """
         Given parameters, get the data from a conventional diagnostic file
         INPUT:
@@ -190,13 +195,15 @@ class Conventional(GSIdiag):
                 analysis_use : if True, will return two sets of data: assimlated
                                (analysis_use_flag=1), and monitored (analysis_use
                                _flag=-1); default = False
-                plvls        : List of pressure levels i.e. [250,500,750,1000]. Will 
-                               return a dictionary with subsetted pressure levels where
-                               data is seperated within those levels:
+                lvls         : List of pressure or height levels i.e. [250,500,750,1000]. List 
+			       must be arranged low to high.  Will return a dictionary with
+                               subsetted pressure or height levels where data is seperated 
+                               within those levels:
 
                                dict = {250-500: <data>,
                                        500-750: <data>,
                                        750-1000: <data>}
+                lvl_type     : lvls definition as 'pressure' or 'height'.  Default is 'pressure'.
 
         OUTPUT:
             data   : requested data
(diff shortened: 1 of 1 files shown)


pyGSI/diags.py
```diff
@@ -156,9 +157,11 @@ def _read_obs(self):
             self.lons = f.variables['Longitude'][:]
             self.lats = f.variables['Latitude'][:]
             self.press = f.variables['Pressure'][:]
+            self.height = f.variables['Height'][:]/1000  	# convert height from meters to km
```
CoryMartin-NOAA, line 160: Why convert from meters to km? Is this just for plotting? If so, suggest converting it later.

pyGSI/diags.py
```diff
@@ -190,13 +194,15 @@ def get_data(self, diag_type, obsid=None, subtype=None, station_id=None, analysi
...
-                               data is seperated within those levels:
+                lvls         : List of pressure or height levels i.e. [250,500,750,1000]. List 
+			       must be arranged low to high.  Will return a dictionary with
+                               subsetted pressure or height levels where data is seperated 
+                               within those levels:
 
                                dict = {250-500: <data>,
                                        500-750: <data>,
                                        750-1000: <data>}
+                lvl_type     : lvls definition as 'pressure' or 'height'.  Default is 'pressure'.
```
CoryMartin-NOAA, line 205: this says default is 'pressure' but lvl_type default is `None`. Suggest changing the function definition so that lvl_type='pressure'

pyGSI/diags.py
```diff
@@ -408,27 +443,46 @@ def _get_idx_conv(self, obsid=None, subtype=None, station_id=None, analysis_use=
 
             return assimilated_idx, monitored_idx
 
-    def get_lat_lon(self, obsid=None, subtype=None, station_id=None, analysis_use=False, plvls=None):
+    def get_lat_lon(self, obsid=None, subtype=None, station_id=None, analysis_use=False, lvls=None, lvl_type=None):
```
CoryMartin-NOAA, line 446: Same comments as above about default lvl_type.

pyGSI/diags.py
```diff
@@ -438,25 +492,25 @@ def get_lat_lon(self, obsid=None, subtype=None, station_id=None, analysis_use=Fa
...
-                        self.press < pressure_list[i+1]))
+                for i, pressure in enumerate(level_list[:-1]):
+                    pres_idx = np.where((self.press > level_list[i]) & (
+                        self.press < level_list[i+1]))
                     valid_idx = np.isin(idx[0], pres_idx[0])
                     pidx = np.where(valid_idx)
 
-                    pressure_lats[pressure] = self.lats[pidx]
-                    pressure_lons[pressure] = self.lons[pidx]
+                    binned_lats[pressure] = self.lats[pidx]
```
CoryMartin-NOAA, line 510: shouldn't this be `binned_lats[low_bound]`?

CoryMartin-NOAA (changes requested):

pyGSI/diags.py
EdwardSafford-NOAA: Yes, it's just for plotting.  Will remove the conversion and implement it in the new vertical plotting routine.

pyGSI/diags.py
EdwardSafford-NOAA: Will do.

pyGSI/diags.py
```diff
@@ -208,21 +214,41 @@ def get_data(self, diag_type, obsid=None, subtype=None, station_id=None, analysi
         self.metadata['Station ID'] = station_id
         self.metadata['Anl Use'] = analysis_use
 
-        if plvls is not None:
-            pressure_list = plvls
-            binned_pressure = {}
+        if lvls is not None:
+            if lvl_type is not None:
+                if lvl_type != 'height' and lvl_type != 'pressure':
+                    print( "WARNING:  input lvl_type ', lvl_type, ' does not match expected values of height or pressure" )
```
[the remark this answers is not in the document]
EdwardSafford-NOAA: Will do.

pyGSI/diags.py
EdwardSafford-NOAA: Will do.

pyGSI/diags.py
EdwardSafford-NOAA: Indeed.  Good catch.

EdwardSafford-NOAA (commented):
I believe I've made all the requested changes.  I managed though to some how mess up the amended commit and so foiled my squash merge.  In cases like this with the GSI repo, I've been asked to cancel the PR and re-do the squash merge.  If you'd like me to do that here I'll make it so.

CoryMartin-NOAA:
@EdwardSafford-NOAA I don't quite understand why we have to do it with GSI, as I can just do a squash merge on GitHub once everything is approved. Let me do another quick review here now.

pyGSI/diags.py
```diff
@@ -208,21 +215,34 @@ def get_data(self, diag_type, obsid=None, subtype=None, station_id=None, analysi
         self.metadata['Station ID'] = station_id
         self.metadata['Anl Use'] = analysis_use
 
-        if plvls is not None:
-            pressure_list = plvls
-            binned_pressure = {}
+        if lvls is not None:
+
+            if lvl_type not in _VALID_LVL_TYPES:
+                raise ValueError('{lvl_type} wrong, use "pressure" or "height" for input lvl_type'.format(lvl_type=repr(lvl_type)))
```
CoryMartin-NOAA, line 221: This is fine and doesn't need to be changed, because I doubt we will have more than pressure or height, but you can do something like:
`raise ValueError(f'{lvl_type} wrong. Valid choices are: {" | ".join(_VALID_LVL_TYPES)}')`

EdwardSafford-NOAA:
Sorry my wording was vague.  I meant when working with the GSI package it was handled as I described, but I wasn't sure canceling and opening a new PR was the preferred method within the PyGSI package.

CoryMartin-NOAA (approved):

CoryMartin-NOAA:
Looks good @EdwardSafford-NOAA , thanks for adding this!

pyGSI/diags.py
EdwardSafford-NOAA: Good idea.  I'll add that with my next set of changes.

Issue #62, opened 2021-09-07
Fix diags.py get_lat_lon()


EdwardSafford-NOAA:
When I added binning by height I missed a needed update to diags.py get_lat_lon().  

The else block beginning at line 542 should have a check on the lvl_type variable and handle 'height' separately from the default value 'pressure'.  There is an example of correct handling of lvl_type within the same function at line 507.

EdwardSafford-NOAA:
Discovered another problem.  Recent changes to the class Convention _read_obs() routine added retrieval of the Prep_QC_Mark from the conventional diag file.  Problem (as reported below) is that gps conventional data helpfully uses Setup_QC_Mark not Prep_QC_Mark.  I'll add that to my fix list.

```
Traceback (most recent call last):
  File "/scratch1/NCEPDEV/da/Edward.Safford/noscrub/PyGSI/scripts/mp_plot_conv_diags.py", line 113, in <module>
    p.map(plotting, work)
  File "/scratch1/NCEPDEV/da/Kevin.Dougherty/anaconda3/envs/PyGSI/lib/python3.8/multiprocessing/pool.py", line 364, in map
    return self._map_async(func, iterable, mapstar, chunksize).get()
  File "/scratch1/NCEPDEV/da/Kevin.Dougherty/anaconda3/envs/PyGSI/lib/python3.8/multiprocessing/pool.py", line 771, in get
    raise self._value
KeyError: 'Prep_QC_Mark'
```
@XiujuanSu-NOAA please consider running the scripts/mp_plot_conv_diags.py as a general test after changing diags.py.

EdwardSafford-NOAA:
Hit this problem immediately when I started work on this issue and am still nowhere as far as understanding why it's now a problem.  I had written a test driver to run the get_data() routine in pyGSI/diags.py but with Su's changes it now produces this error which I'm at a loss to understand:
```
File "test_driver.py", line 102, in plotting
    data = diag.get_data( diag_type, obsid=obsid,
  File "/scratch1/NCEPDEV/da/Edward.Safford/noscrub/PyGSI/pyGSI/diags.py", line 322, in get_data
    hght_idx = np.where((self.height > level_list[i]) & (
TypeError: '>' not supported between instances of 'MaskedArray' and 'str'
```

CoryMartin-NOAA:
@EdwardSafford-NOAA can you print out both self.height and level_list[i]? Seems like one of them (level_list?) might be a string and not a float/int value?

[...]

PR #63, opened 2021-09-08
Fix omissions in pyGSI/diags.py and plot_diags.py

EdwardSafford-NOAA:
Fix items described in #62 including
  

- get_lat_lon()
- read_obs() to fix the gps use of "Setup_QC_Mark" instead of "Prep_QC_Mark" like the rest of the data types
- add checks to diags.py for valid binning and observation types
- add missing analysis type 'rejected' to both diags.py and plot_diags.py


EdwardSafford-NOAA:
Something is wrong - there are no changed files.  I'm going to close this and try again.

PR #66, opened 2021-09-08
Fix issue #62, missed items in diags.py

EdwardSafford-NOAA:
Fix items described in #62 including

- get_lat_lon()
- read_obs() to fix the gps use of "Setup_QC_Mark" instead of "Prep_QC_Mark" like the rest of the data types
- add checks to diags.py for valid binning and observation types
- add missing analysis type 'rejected' to both diags.py and plot_diags.py

### DIFF
--- a/pyGSI/diags.py
+++ b/pyGSI/diags.py
@@ -9,6 +9,7 @@ from datetime import datetime
 from pathlib import Path
 
 _VALID_LVL_TYPES = ["pressure", "height"]
+_VALID_DIAG_TYPES = ["omf", "o-f", "omb", "o-b", "oma", "o-a", "observation", "hofx"]
 
 class GSIdiag:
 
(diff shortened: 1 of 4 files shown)


pyGSI/diags.py
```diff
@@ -224,15 +228,18 @@ def get_data(self, diag_type, obsid=None, subtype=None, station_id=None, analysi
         OUTPUT:
             data   : requested data
         """
-        
+       
+        if diag_type not in _VALID_DIAG_TYPES:
+            raise ValueError(f'{diag_type} wrong. Valid choices are: {" | ".join(_VALID_DIAG_TYPES)}')
```
CoryMartin-NOAA, line 233: this will print out `blah wrong. valid choices are...`. Should it be more grammatically correct, something like `blah is not a valid diag_type. Valid choices are:...`

CoryMartin-NOAA (approved):
Thank you @EdwardSafford-NOAA for these fixes. Looks good to me. One comment but I see it is a copy of previous lines so not required to change it here.

PR #68, opened 2021-09-09
Update `diags.py`

kevindougherty-noaa:
Using a new approach to use a multidimensional dataframe that can use input arguments to select a slice of the dataframe and make indexing much easier and cleaner.

Each commit will take one step at a time so reviewers can see the progress and is not one large change. This PR will remain a draft, and then when I replicate all of diags.py, I will add implemented changes to diags.py and ask for a final review.

This first commit works on `Conventional()` and how to grab and index data. This should be the largest change in my steps.


kevindougherty-noaa:
@CoryMartin-NOAA made a good point that I should make the changes in diags.py rather than a new file so you can see the changes easier. No need to review `diags_v2.py`. I will remove it next commit.

EdwardSafford-NOAA (commented):
Per your comments I only looked at the changes to diags.py and noted a couple of questions.  I'm looking forward to working with pandas -- looks like it's going to be a bit easier to use.

pyGSI/diags.py
```diff
@@ -40,203 +42,164 @@ def __init__(self, path):
...
+
+                # Grab variables with only 'nobs' dimension
+                elif len(f.variables[var].shape) == 1:
+                    df_dict[var] = f.variables[var][:]
+
+        # Create pandas dataframe from dict
+        df = pd.DataFrame(df_dict)
+
+        # Creates multidimensional indexed dataframe
+        indices = ['Station_ID', 'Observation_Class', 'Observation_Type',
```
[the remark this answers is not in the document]
kevindougherty-noaa: Got it. Will be in the next commit.

kevindougherty-noaa:
@CoryMartin-NOAA @EdwardSafford-NOAA @XiujuanSu-NOAA Updates to `Conventional()` have been addressed. `Radiance` and `Ozone` will follow similar approach if reviews are good with all of you.

EdwardSafford-NOAA (commented):
Looks good to me.  Thanks for cleaning up the phrasing on the checks to _VALID_*TYPES.

pyGSI/diags_v2.py
```diff
@@ -0,0 +1,334 @@
+import os
```
CoryMartin-NOAA, line 1: Can't this file be completely removed from the PR? Or am I missing something?

pyGSI/diags.py
```diff
@@ -1,15 +1,17 @@
...
 import os
+import pandas as pd
 import numpy as np
 from netCDF4 import Dataset
 from datetime import datetime
 from pathlib import Path
 
 _VALID_LVL_TYPES = ["pressure", "height"]
-_VALID_DIAG_TYPES = ["omf", "o-f", "omb", "o-b", "oma", "o-a", "observation", "hofx"]
+_VALID_CONV_DIAG_TYPES = ["omf", "oma", "observation", "hofx"]
```
CoryMartin-NOAA, line 9: Previously you had `o-f` and `omf`, now it is just `omf`, I'm okay with that, but for backwards compatibility/users, can we keep this as just `omf` but sanitize the input when this is called to remove `-` and replace it with `m`?

pyGSI/diags.py
```diff
@@ -40,327 +42,187 @@ def __init__(self, path):
...
-                self.v_o = f.variables['v_Observation'][:]
-
-                self.u_omf = f.variables['u_Obs_Minus_Forecast_adjusted'][:]
-                self.v_omf = f.variables['v_Obs_Minus_Forecast_adjusted'][:]
-
-            else:
-                self.o = f.variables['Observation'][:]
-                self.omf = f.variables['Obs_Minus_Forecast_adjusted'][:]
+            for var in f.variables:
+                # Station_ID and Observatio_Class variables need
```
CoryMartin-NOAA, line 102: typo, `Observation_Class`

pyGSI/diags.py
```diff
@@ -40,327 +42,187 @@ def __init__(self, path):
...
+        df.columns = df.columns.str.lower()
+        if self.variable == 'uv':
+            df = df.rename(columns={
+                'u_obs_minus_forecast_unadjusted': 'u_omf_unadjusted',
+                'u_obs_minus_forecast_adjusted': 'u_omf_adjusted',
+                'v_obs_minus_forecast_unadjusted': 'v_omf_unadjusted',
+                'v_obs_minus_forecast_adjusted': 'v_omf_adjusted'
+                })
+            # Create hofx columns
+            df['u_hofx_unadjusted'] = df['u_observation'] - \
```
CoryMartin-NOAA, line 136: could do this in a double loop
```
for i in 'u', 'v':
  for j in 'unadjusted', 'adjusted':
```

but not sure how much cleaner that is. Might be harder to read/understand.

pyGSI/diags.py
```diff
@@ -40,327 +42,187 @@ def __init__(self, path):
...
+                   low to high.
+            lvl_type : (str; default='pressure') lvls definition as
+                       'pressure' or 'height'.
+            bias_correction : (bool; default=True) If True, will return bias
+                              corrected data.
+        Returns:
+            data : requested indexed data
+        """
+
+        if diag_type not in _VALID_CONV_DIAG_TYPES:
```
CoryMartin-NOAA, line 187: This is where I would propose sanitizing the `diag_type`, if you wish to do that.

pyGSI/diags.py
```diff
@@ -386,231 +249,239 @@ def get_data(self, diag_type, obsid=None, subtype=None, station_id=None, analysi
...
-                lons = {'assimilated': self.lons[assimilated_idx],
-                        'rejected': self.lons[rejected_idx],
-                        'monitored': self.lons[monitored_idx]}
-                return lats, lons
-            else:
-                idx = self._get_idx_conv(
-                    obsid, subtype, station_id, analysis_use)
-                return self.lats[idx], self.lons[idx]
+            df = df.query(
+                f'(Height <= {high_bound}) and (Height > {low_bound})')
```
CoryMartin-NOAA, line 438: I think before, pressure was <, >= and height was =<, >, is that correct? @EdwardSafford-NOAA should they both be consistent in code, or consistent in physical space (i.e. height should be opposite of pressure because of the inverse relation)

CoryMartin-NOAA (commented):
A few comments, but mostly looks good. Thanks @kevindougherty-noaa for switching to pandas

pyGSI/diags.py
EdwardSafford-NOAA: @CoryMartin-NOAA yes you are correct, but honestly I'm not sure it matters much.  I think the initial idea was consistency in physical space, but I have to ask @XiujuanSu-NOAA -- was consistency in physical space the intent?

CoryMartin-NOAA: I think it should be consistent in physical space myself, if you did:
`[1100, 500, 0]` in hPa space, I would want it to be roughly the same as
`[0,5500,99999]` in m space

pyGSI/diags.py
kevindougherty-noaa: I think I would prefer to take Rahul's original suggestion and just have the set inputs. If it's wrong, the user is notified right away with the correct inputs with the value error.

pyGSI/diags_v2.py
kevindougherty-noaa: Right, I will remove in next commit.

kevindougherty-noaa:
I have added updates to your comments and have added the new `Radiance()` class.

pyGSI/diags.py
```diff
@@ -1,15 +1,17 @@
...
+import pandas as pd
 import numpy as np
 from netCDF4 import Dataset
 from datetime import datetime
 from pathlib import Path
 
 _VALID_LVL_TYPES = ["pressure", "height"]
-_VALID_DIAG_TYPES = ["omf", "o-f", "omb", "o-b", "oma", "o-a", "observation", "hofx"]
+_VALID_CONV_DIAG_TYPES = ["omf", "oma", "observation", "hofx"]
+_VALID_RADIANCE_DIAG_TYPES = ["omf", "oma", "observation", "hofx",
```
CoryMartin-NOAA, line 10: Are all of the `fraction` variables properly supported now in this latest commit?

kevindougherty-noaa: Yes they should be. I can update that list further for the ones that I missed.

kevindougherty-noaa:
@CoryMartin-NOAA @EdwardSafford-NOAA @XiujuanSu-NOAA I have updated all classes in `diags.py`. Please let me know if you catch any typos or things I missed. I am going to stress test this script to ensure every possible scenario works.

CoryMartin-NOAA:
@kevindougherty-noaa let us know whenever you are ready for a thorough/final review by changing this from a draft pull request by clicking 'ready for review'

XiujuanSu-NOAA:
I only familiar with conventional diagnostic type such as omf oma or omb
and observation

Su

kevindougherty-noaa:
@CoryMartin-NOAA @EdwardSafford-NOAA I ran some stress tests and made some changes. Everything should be good to go for a full review.

pyGSI/diags.py
```diff
@@ -40,327 +43,344 @@ def __init__(self, path):
...
-                        u, v = self.query_diag_type(diag_type, pidx)
+    def _select_conv(self, obsid=None, subtype=None, station_id=None,
+                     analysis_use=False):
+        """
+        Given parameters, multidimensional dataframe is indexed
+        to only include selected locations from a conventional
+        diagnostic file.
+
+        Args:
+            obsid : (list of ints; default=None) observation measurement
```
CoryMartin-NOAA, line 276: "measurement ID number"? do you mean `120 = sondes` for example? Suggest changing this to something like `observation type number`

pyGSI/diags.py
```diff
@@ -40,327 +43,344 @@ def __init__(self, path):
...
+            subtype : (list of ints; optional; default=None) observation
+                      measurement ID subtype number, default=None
+            station_id : (list of str; optional; default=None)
+                         station id, default=None
+            analysis_use : (bool; default=False) if True, will return
+                           three sets of data:
+                           assimilated (analysis_use_flag=1, qc<7),
+                           rejected (analysis_use_flag=-1, qc<8),
+                           monitored (analysis_use_flag=-1, qc>7)
+            lvls : (list type; default=None) List of pressure or height
```
CoryMartin-NOAA, line 178: I think you need here some documentation about the `<`, `>`, etc work, like which is inclusive for pressure and height, etc.

pyGSI/diags.py
```diff
@@ -369,259 +389,260 @@ def get_data(self, diag_type, obsid=None, subtype=None, station_id=None, analysi
...
-                idxobs = self.o_type
-                valid_obs_idx = np.isin(idxobs, obsid)
+    def get_pressure(self, obsid=None, subtype=None, station_id=None,
+                     analysis_use=False):
+        """
+        Grabs indexed pressure data.
+
+        Args:
+            obsid : (list of ints; optional; default=None) observation
+                    measurement ID number; default=None
```
CoryMartin-NOAA, line 455: again, measurement ID number sounds too specific IMO, and needs to reflect it is the data type instead

pyGSI/diags.py
```diff
@@ -633,164 +654,241 @@ def __str__(self):
...
+
+            df = df.iloc[indx]
+
+            # remove obs where inverse obs error is zero
+            if errcheck and 0 in qcflag:
+                indx = df.index.get_level_values(idx_col) == ''
+
+                # Grab index where inverse ob error is not zero
+                err_indx = np.isin(
+                    df['inverse_observation_error'], 0, invert=True)
```
CoryMartin-NOAA, line 802: Did you test this? is it actually 0 or a really small number in inverse_observation_error?

CoryMartin-NOAA (commented):
a couple of small comments/nitpicks, otherwise I think it looks good.

XiujuanSu-NOAA:
I agree with Cory, instead of instrument ID, I think observation type or
prebufr data type is better since these numbers (120 130..) are defined in
prepbufr file.

Su

pyGSI/diags.py
kevindougherty-noaa: @CoryMartin-NOAA It is actually zero. I just double checked.

pyGSI/diags.py
```diff
@@ -369,259 +395,266 @@ def get_data(self, diag_type, obsid=None, subtype=None, station_id=None, analysi
...
+        Args:
+            obsid : (list of ints; optional; default=None) observation
+                    type ID number; default=None
+            subtype : (list of ints; optional; default=None) observation
+                      measurement ID subtype number, default=None
+            station_id : (list of str; optional; default=None)
+                         station id, default=None
+            analysis_use : (bool; default=False) if True, will return
+                           three sets of data:
+                           assimlated (analysis_use_flag=1, qc<7),
```
EdwardSafford-NOAA, line 468: Nitpicking -- typo with assimilated.

EdwardSafford-NOAA (commented):
Looks good to me.

pyGSI/diags.py
kevindougherty-noaa: Nitpicking makes good code! Good catch, thank you.

EdwardSafford-NOAA (approved):
Looks good to me.

CoryMartin-NOAA (approved):
Looks good to me, thanks @kevindougherty-noaa

PR #70, opened 2021-09-23
Add use flag check

kevindougherty-noaa:
`Radiance()` had `use_flag` as an input, but did not address it. This PR adds that and addresses some other small details needed for plotting routine.

### DIFF
--- a/pyGSI/diags.py
+++ b/pyGSI/diags.py
@@ -789,7 +789,7 @@ class Radiance(GSIdiag):
         if use_flag:
             use_flag_indx = np.where(self.chan_info['use_flag'] == 1)
             use_flag_channel = self.chan_info[
-                'sensor_chan']\[use_flag_indx].tolist()
+                'sensor_chan'][use_flag_indx].tolist()
 
             self.metadata['Channels'] = use_flag_channel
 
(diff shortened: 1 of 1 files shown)


pyGSI/diags.py
```diff
@@ -750,9 +750,17 @@ def get_data(self, diag_type, channel=None, qcflag=None, use_flag=False,
                               'Valid choices are: '
                               f'{" | ".join(_VALID_RADIANCE_DIAG_TYPES)}'))
 
+        self.metadata['Variable'] = 'brightness_temperature'
```
EdwardSafford-NOAA, line 753: Unrelated to your changes, please check line 727.  I think there is a cut and paste error -- this method is working with radiance diag files, not conventional diag files, right?

EdwardSafford-NOAA (commented):
Looks ok to me.  Just had the one, unrelated question.

CoryMartin-NOAA (dismissed):

pyGSI/diags.py
kevindougherty-noaa: @EdwardSafford-NOAA Good catch. I just fixed it and will be included in the PR.

CoryMartin-NOAA (approved):

PR #74, opened 2021-09-28
Bug fix for len function in diags.py

ADCollard:
The len() function in diags.py used the "lats" variable which is no longer defined.  This is a simple fix to reinstate the len() functionality.


kevindougherty-noaa (approved):
Looks good. Don't mind the pycodestyle not passing, that is due to other files. Adding new workflow management to the repository!

CoryMartin-NOAA (approved):

Issue #75, opened 2021-09-29
Add feature to return dataframe rather than numpy array


kevindougherty-noaa:
In get data, add an input as `return_df` as a boolean. If `True`, return the dataframe.

aerorahul:
Is there a reason, not to **always** return a dataframe?
What is the advantage of returning just the variables?

kevindougherty-noaa:
@aerorahul You bring up a good question. I think the original idea when creating this interface was to be made for those not familiar with python. That way, all you need is two lines to get an array of a specific variable that you want (i.e. omf, oma, observation, hofx). Returning just the dataframe would remove a lot of lines from `diags.py`. It may require some changes in other scripts, but I am indifferent about keeping the current way or changing to just return the dataframe. What do you think? A decision should be made before I start diving in with developing regional DA monitoring scripts any further.

PR #76, opened 2021-10-05
Feature/return df

kevindougherty-noaa:
Addresses #75

### DIFF
--- a/pyGSI/diags.py
+++ b/pyGSI/diags.py
@@ -277,7 +277,6 @@ class Conventional(GSIdiag):
                     return self._query_diag_type(
                             indexed_df, diag_type, bias_correction)
 
-
     def _select_conv(self, obsid=None, subtype=None, station_id=None,
                      analysis_use=False):
         """
--- a/pyGSI/diags.py
+++ b/pyGSI/diags.py
@@ -643,6 +643,10 @@ class Conventional(GSIdiag):
 
                 return lats, lons
 
+	def return_df(self):
+		"""Returns working dataframe."""
+		return self.data_df
+
 
 class Radiance(GSIdiag):
 
@@ -907,6 +911,10 @@ class Radiance(GSIdiag):
 
         return lats, lons
 
+	def return_df(self):
+        """Returns working dataframe."""
+        return self.data_df
+
 
 class Ozone(GSIdiag):
 
@@ -1131,3 +1139,8 @@ class Ozone(GSIdiag):
                     lons_dict[p] = indexed_df['longitude'].to_numpy()
 
         return lats_dict, lons_dict
+
+	def return_df(self):
+        """Returns working dataframe."""
+        return self.data_df
+
--- a/pyGSI/diags.py
+++ b/pyGSI/diags.py
@@ -157,7 +157,7 @@ class Conventional(GSIdiag):
 
     def get_data(self, diag_type, obsid=None, subtype=None, station_id=None,
                  analysis_use=False, lvls=None, lvl_type='pressure',
-                 bias_correction=True):
+                 bias_correction=True, return_df=False):
         """
         Given parameters, get the data from a conventional diagnostic file
 
@@ -188,6 +188,8 @@ class Conventional(GSIdiag):
                        'pressure' or 'height'.
             bias_correction : (bool; default=True) If True, will return bias
                               corrected data.
+			return_df : (bool; defualt=False) If True, will return entire
+						indexed dataframe rather than numpy array of data
         Returns:
             data : requested indexed data
         """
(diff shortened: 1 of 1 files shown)


kevindougherty-noaa:
@CoryMartin-NOAA still need to go through the other files for `pycodestyle` to pass, however `diags.py` passes which is the only file I changed.

CoryMartin-NOAA:
@kevindougherty-noaa can you explain to @ADCollard and me why this PR is needed? I'm sure it is a method vs attribute thing but it's not clear to me at the moment.

kevindougherty-noaa:
@CoryMartin-NOAA @ADCollard Sure thing. In our regional DA meeting last week, Ting and Shun mentioned interest in being able to have the entire dataframe returned. I could be mistaken and maybe they meant returning the dataframe after being indexed.

pyGSI/diags.py
```diff
@@ -643,6 +643,10 @@ def get_lat_lon(self, obsid=None, subtype=None, station_id=None,
 
                 return lats, lons
 
+    def return_df(self):
+        """Returns working dataframe."""
+        return self.data_df
+
```
aerorahul: Why can't you add this method in `Class GSIdiag` since that is the class from which all other inherit from?

aerorahul (changes requested):
See comment.

CoryMartin-NOAA:
@kevindougherty-noaa yes I remember that, so I guess you're saying this is so one can have numpy arrays if they want, and a Pandas dataframe through the `.return_df()` method? I thought the issue you linked would basically imply that the read method would have an argument like `read_data(return_df=True)` and then:
```
if return_df:
    return self.df
else:
   return data
```
or whatever the above pseudo code should be

kevindougherty-noaa:
@CoryMartin-NOAA you are right, that is what should be done. I will fix.

pyGSI/diags.py
```diff
@@ -224,52 +227,64 @@ def get_data(self, diag_type, obsid=None, subtype=None, station_id=None,
...
-                    monitored_data = self._query_diag_type(
-                        monitored_df, diag_type, bias_correction)
-
-                    data = {'assimilated': assimilated_data,
-                            'rejected': rejected_data,
-                            'monitored': monitored_data
-                            }
-
-                    return data
+				if return_df:
```
CoryMartin-NOAA: @kevindougherty-noaa is this indentation correct?

pyGSI/diags.py
```diff
@@ -224,52 +227,64 @@ def get_data(self, diag_type, obsid=None, subtype=None, station_id=None,
...
-
-                    u = {'assimilated': u_assimilated,
-                         'rejected': u_rejected,
-                         'monitored': u_monitored}
-                    v = {'assimilated': v_assimilated,
-                         'rejected': v_rejected,
-                         'monitored': v_monitored}
-
-                    return u, v
+                if return_df:
```
CoryMartin-NOAA, line 230: so for non data frame, it returns a dict of dicts, for u,v. For the data frame case, are both u and v inside the data frame?

kevindougherty-noaa: Yes, both u and v are in the dataframe. All the variables will be included.

pyGSI/diags.py
```diff
@@ -224,52 +227,64 @@ def get_data(self, diag_type, obsid=None, subtype=None, station_id=None,
...
-                        indexed_df, diag_type, bias_correction)
+                    if self.variable == 'uv':
+                        u, v = self._query_diag_type(
+                            indexed_df, diag_type, bias_correction)
+
+                        return u, v
+
+                    else:
+                        data = self._query_diag_type(
+                            indexed_df, diag_type, bias_correction)
```
aerorahul:
```suggestion
                    return self._query_diag_type(
                            indexed_df, diag_type, bias_correction)
```

pyGSI/diags.py
```diff
@@ -224,52 +227,64 @@ def get_data(self, diag_type, obsid=None, subtype=None, station_id=None,
...
 
-                    return data
+                        return data
```
aerorahul:
```suggestion
```

aerorahul (changes requested):
Just a quick glance. 
See if you notice these as well.

pyGSI/diags.py
CoryMartin-NOAA: This is a good suggestion

kevindougherty-noaa: I like it, thanks @aerorahul.

kevindougherty-noaa:
Based on conversations with @CoryMartin-NOAA and @aerorahul, I am going to close this PR and develop `diags.py` to only return dataframes rather than specific variables. It should cut out a good amount of code.

PR #77, opened 2021-10-06
Change `diags.py` to only return dataframes

kevindougherty-noaa:
Discussions were had to change `diags.py` to only return dataframes. This reduced overall code, however multiple scripts will need to be changed after the merge of this PR. Addresses #75


pyGSI/diags.py
```diff
@@ -6,12 +6,6 @@
 from pathlib import Path
 
 _VALID_LVL_TYPES = ["pressure", "height"]
-_VALID_CONV_DIAG_TYPES = ["omf", "oma", "observation", "hofx"]
```
CoryMartin-NOAA, line 9: how does one request hofx? do they need to manually grab `Observation` and `Observation_Minus_Forecast_Adjusted`?

kevindougherty-noaa: @CoryMartin-NOAA an hofx column is created when the dataframe is developed for all three classes. For example, see lines 104 and 114.

pyGSI/diags.py
```diff
@@ -840,72 +533,92 @@ def _select_radiance(self, channel=None, qcflag=None, use_flag=False,
...
+            err_indx = np.isin(
+                good_channels_df['inverse_observation_error'], 0, invert=True)
+            assimilated_indx = np.ma.logical_or(assimilated_indx, err_indx)
+
+            assimilated_df = good_channels_df.iloc[assimilated_indx]
 
-    def _get_data_special(self, diag_type, channel, qcflag, use_flag,
-                          separate_channels, separate_qc, errcheck,
-                          bias_correction):
+            # Get rejected data where QC != 1
```
CoryMartin-NOAA, line 600: Typo, also what about monitored? I think rejected needs to be qc > 0? Or does the 'good channels' handle this?

CoryMartin-NOAA: Do assim + monitored + rejected for each channel combined result in the right total?

CoryMartin-NOAA (commented):
Just saw a minor typo. However, don't want to approve/merge this right now as won't that break the develop branch? This needs to be coordinated with changes to the examples/scripts I think.

CoryMartin-NOAA:
@kevindougherty-noaa we can create a working 'tag' in GitHub for people to use if you want to 'break' develop temporarily though.

pyGSI/diags.py
kevindougherty-noaa: What I was told by Xiaoyan and Haixia is that:

- assimilated data : QC_Flag == 0 and inv_ob_error != 1
- rejected data : QC_Flag != 0
- monitored data : use_flag != 1

So monitored is included just above that. When I add the three, they are equal to the combined total.

kevindougherty-noaa:
@CoryMartin-NOAA @aerorahul changes to `diags.py` and the respective scripts have been updated.

pyGSI/diags.py
```diff
@@ -213,66 +171,28 @@ def get_data(self, diag_type, obsid=None, subtype=None, station_id=None,
...
-                else:
-                    data = self._query_diag_type(
-                        indexed_df, diag_type, bias_correction)
-
-                    return data
+            return indexed_df
 
-    def _select_conv(self, obsid=None, subtype=None, station_id=None,
-                     analysis_use=False):
+    def _select_conv(self, obsid, subtype, station_id, analysis_use=False):
```
CoryMartin-NOAA: Why now is just analysis_use with a default? Shouldn't it be all or nothing for this internal method?

CoryMartin-NOAA: Also the documentation below needs to be updated as appropriate for if there are/aren't defaults.

pyGSI/diags.py
```diff
@@ -840,72 +563,110 @@ def _select_radiance(self, channel=None, qcflag=None, use_flag=False,
 
                 df = df.iloc[indx]
 
-        return df
+        if analysis_use:
+            # Separate into 3 dataframes; assimilated, rejected, and monitored
+            # First step, separate use_flag=1 and useflag!=1
+            good_use_flag_indx = np.where(self.chan_info['use_flag'] == 1)
+            bad_use_flag_indx = np.where(self.chan_info['use_flag'] != 1)
```
CoryMartin-NOAA: I suggest not saying this is 'bad' per se, but rather 'monitored' Maybe have `used_indx` and `monitor_indx` or something like that?

pyGSI/plot_diags.py
```diff
@@ -157,36 +158,39 @@ def _get_labels(metadata):
...
-        save_file = save_file + '%s' % '_'.join(
-            str(x) for x in metadata['ObsID Name'])
+        save_file = save_file + '_%s' % '_'.join(
+            str(x).replace(" ", "_") for x in metadata['ObsID'])
 
     elif metadata['Diag File Type'] == 'radiance':
         if metadata['Channels'] == 'All Channels':
             title = title + '\nAll Channels'
-            save_file = save_file + 'All_Channels'
+            save_file = save_file + '_All_Channels'
```
CoryMartin-NOAA: nit picking here, but below is `channels` and here it is `All_Channels`, this should probably be lowercase?

scripts/mp_plot_sat_diags.py
```diff
@@ -18,20 +17,54 @@ def plotting(sat_config):
     diag_type = sat_config['radiance input']['data type'][0].lower()
     channel = sat_config['radiance input']['channel']
     qcflag = sat_config['radiance input']['qc flag']
+#     analysis_use = sat_config['radiance input']['analysis use'][0]
```
CoryMartin-NOAA: can this line that is commented out be removed? or is it a mistake?

CoryMartin-NOAA (commented):
Few comments but overall looks good.

scripts/mp_plot_sat_diags.py
kevindougherty-noaa: It should be uncommented. The current sample yamls do not include `analysis use` as an option and forgot to uncomment.

CoryMartin-NOAA (approved):

PR #80, opened 2021-10-12
Add list info methods for respective classes

kevindougherty-noaa:
When method is called, will print all values for specific inputs based on the class you are using. i.e. `Conventional` has `list_obsids()` and `list_stationids()`, `Radiance` has `list_channels()` and `list_qcflags()` and `Ozone` has `list_pressures()`

### DIFF
--- a/pyGSI/diags.py
+++ b/pyGSI/diags.py
@@ -330,6 +330,26 @@ class Conventional(GSIdiag):
 
         return df
 
+    def list_obsids(self):
+        """
+        Prints all the unique observation ids in the diagnostic file.
+        """
+        df = self.data_df.reset_index()
+        obsids = sorted(df.Observation_Type.unique())
+
+        print('Observation IDs in this diagnostic file include:\n'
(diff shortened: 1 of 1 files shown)


pyGSI/diags.py
```diff
@@ -330,6 +330,26 @@ def _select_levels(self, df, low_bound, high_bound, lvl_type):
 
         return df
 
+    def list_obsids(self):
```
CoryMartin-NOAA, line 339: One comment:
Would there be utility in having these stored as (example):
`self.obsids`
and then the method would print
`join([str(x) for x in self.obsids])}')`
So that it is easy to loop over all unique obs, etc.

kevindougherty-noaa: Oh definitely. Good catch.

CoryMartin-NOAA: could we do something like:

`diag.obsids.list()` to achieve this? is that easily done?

CoryMartin-NOAA (approved):

PR #83, opened 2021-10-13
Add script to plot QC Flags on map

kevindougherty-noaa:
Creates a 2D map that plots the location of the qc flags. Addresses number 2.), 4th bullet in #84. An example is shown here:

image

Need to create a yaml and driver script to run.


LAMDA/no_data_plots.py
```diff
@@ -0,0 +1,42 @@
+import numpy as np
+import matplotlib.pyplot as plt
+import plot_features as features
+
+
+def no_data_map(plotmap, domain, metadata):
```
CoryMartin-NOAA, line 6: I think this should be moved to a more generic part of `emcpy` as there may be other uses for this outside of LAMDA

LAMDA/map_qc_flags.py
```diff
@@ -0,0 +1,120 @@
...
+                       assimilated (QC_Flag=0, inv_observation_error!=0),
+                       rejected (QC_Flag!=0),
+                       monitored (use_flag!=1)
+        domain : (str; default='conus') domain in which to plot data
+        projection : (str; default='plcarr') projection of map to plot
+                     data
+        plotdir : (str; default='./') path to where figures should be saved
+    """
+
+    diag = Radiance(inputfile)
```
CoryMartin-NOAA: Can we make this generic for all obs types? What if we want to plot QC values of conventional data (say `satwinds`)?  Will it work to just pass the diag object to this function, or do we need to handle all three classes here?

CoryMartin-NOAA (commented):
Two comments which will probably require more work than what is possible before your 2 days of PTO and weekend :-)

LAMDA/map_qc_flags.py
kevindougherty-noaa: That is a good question. To my understanding, there are no qc flags for conventional data, only assimilated, rejected and monitored data. So this script is pretty specific to just `Radiance` data.

CoryMartin-NOAA: There are QC flags for conventional data for sure, I think there are two `Setup_QC_Mark` and `Prep_QC_Mark` but am not totally sure of what should be used (probably Setup)

LAMDA/no_data_plots.py
kevindougherty-noaa: I am going to move it up one directory because it could be useful in other parts of PyGSI as I plan to add other examples that already exist in `pyGSI/plot_diags.py`

PR #87, opened 2021-10-19
Updates to yaml creating scripts for better future multiprocessing techniques

kevindougherty-noaa:


ush/convinfo2yaml.py
```diff
@@ -40,67 +41,79 @@ def main(config):
...
-        # first get filename and verify it exists
+    unique_obtypes = np.unique(obtype)
+
+    for ob in unique_obtypes:
+        yamlout = {'diagnostic': {}}
+
+        obindx = np.where(np.array(obtype) == ob)
         diagfile = os.path.join(f"{config['diagdir']}",
-                                (f"diag_conv_{iobtype}_{config['loop']}"
+                                (f"diag_conv_{ob}_{config['loop']}"
```
CoryMartin-NOAA, line 58: we need to verify this is the same naming convention for global and regional

CoryMartin-NOAA (approved):

ush/convinfo2yaml.py
kevindougherty-noaa: Noted.

PR #88, opened 2021-10-21
LAMDA Workflow Script

kevindougherty-noaa:
This has made some updates to the development of conventional, radiance, and ozone yaml generating scripts as well as an approach to using python multiprocessing to generate plots.


LAMDA/map_qc_flags.py
```diff
@@ -3,10 +3,14 @@
 import os
 from pathlib import Path
 import matplotlib.pyplot as plt
+import sys
+sys.path.append('/scratch1/NCEPDEV/da/Kevin.Dougherty/PyGSI/LAMDA')
```
CoryMartin-NOAA: Don't put these sys.path.appends in the repo.

LAMDA/plot_features.py
```diff
@@ -2,6 +2,9 @@
 import numpy as np
 from textwrap import TextWrapper
 import matplotlib.pyplot as plt
+import sys
+sys.path.append('/scratch1/NCEPDEV/da/Kevin.Dougherty/emcpy/src/')
```
CoryMartin-NOAA: Same comment.

LAMDA/scripts/LAMDA_mapping_workflow.py
```diff
@@ -0,0 +1,124 @@
+import yaml
+import numpy as np
+from multiprocessing import Pool
+from functools import partial
+import itertools
+import argparse
+import sys
+sys.path.append('/scratch1/NCEPDEV/da/Kevin.Dougherty/PyGSI/')
```
CoryMartin-NOAA: again

LAMDA/scripts/LAMDA_mapping_workflow.py
```diff
@@ -0,0 +1,124 @@
...
+    ap.add_argument("-d", "--data_yaml",
+                    help="Path to yaml file with diag data")
+    ap.add_argument("-p", "--plotting_yaml",
+                    help="Path to yaml file with plotting info")
+    ap.add_argument("-o", "--outdir",
+                    help="Out directory where files will be saved")
+
+    myargs = ap.parse_args()
+
+    nprocs = int(myargs.nprocs) if myargs.nprocs else 1
```
CoryMartin-NOAA: can you not instead make the add_argument have type int and default of 1?

LAMDA/scripts/LAMDA_stats_workflow.py
```diff
@@ -0,0 +1,122 @@
...
+if __name__ == '__main__':
+    # Parse command line
+    ap = argparse.ArgumentParser()
+    ap.add_argument("-n", "--nprocs",
+                    help="Number of tasks/processors for multiprocessing")
+    ap.add_argument("-y", "--yaml_input",
+                    help="Path to yaml file")
+
+    myargs = ap.parse_args()
+    nprocs = int(myargs.nprocs) if myargs.nprocs else 1
```
CoryMartin-NOAA: Same comment as on the other script.

CoryMartin-NOAA (commented):
Some quick initial comments.

LAMDA/map_qc_flags.py
```diff
@@ -19,7 +23,7 @@ def _create_map_qc(df, qc_unique, domain, projection,
     # Loop through unique QC Flags to create plot objects
     for i, flag in enumerate(qc_unique):
         if metadata['Diag File Type'] == 'conventional':
-            indx = data_df.index[df['prep_qc_mark'] == flag]
+            indx = df.index[df['prep_qc_mark'] == flag]
```
CoryMartin-NOAA, line 23: One thing that we will need to pay attention to that I think I've neglected to mention is `setup_qc_mark` for temperature. It tells us if it is virtual or sensible temperature.

LAMDA/map_qc_flags.py
```diff
@@ -67,73 +71,76 @@ def _create_map_qc(df, qc_unique, domain, projection,
...
-                     data
-        plotdir : (str; default='./') path to where figures should be saved
+        config : (dict) configuration file that includes the
+                 appropriate inputs based on file type (i.e.
+                 conventional or radiance data)
     """
 
     # Get filename to determing what the file type is
-    filename = os.path.splitext(Path(inputfile).stem)[0]
+    filename = os.path.splitext(Path(config['diag dir']).stem)[0]
```
CoryMartin-NOAA: is `diag dir` then correct here or should it be `diag file`? Does this point to a directory or a full .nc4 file path?

LAMDA/plot_features.py
```diff
@@ -232,12 +230,12 @@ def get_labels(metadata):
     if metadata['Diag File Type'] == 'conventional':
         title = title + '\n%s' % '\n'.join(metadata['ObsID Name'])
         save_file = save_file + '%s' % '_'.join(
-            str(x) for x in metadata['ObsID Name'])
+            str(x) for x in metadata['ObsID'])
```
CoryMartin-NOAA, line 230: How many are expected to be in this list?

LAMDA/scripts/LAMDA_stats_workflow.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def concatenate_dfs(files, variable, cycles):
+    """
+    Reads in list of files and creates a concatenated
+    dataframe of all cycles.
+    """
+    dfs = []
+    for i, file in enumerate(files):
+        gdas = GSIstat(file, cycles[i])
```
CoryMartin-NOAA: Jacob will not like us calling it `gdas` here since that is Global Data Assimilation System  :-) better choose something like `gsi` or `gstat`

LAMDA/scripts/LAMDA_stats_workflow.py
```diff
@@ -0,0 +1,122 @@
...
+
+    return concatenated_df
+
+
+def plotting_func(config, data_dict, outdir):
+
+    variable = config[0]
+    plot_type = config[-1]
+
+    for tm in np.arange(6):
```
CoryMartin-NOAA: isn't this just going to be 0,1,2,3,4,5? but we need tm06

LAMDA/scripts/LAMDA_stats_workflow.py
```diff
@@ -0,0 +1,122 @@
...
+        data_dict[cycle]['fits2'] = sorted(fits2)
+
+    # Create work list
+    if 'minimization' in plot_types:
+        plot_types.remove('minimization')
+        work_list = list(itertools.product(variables, plot_types))
+        work_list.append(('cost', 'minimization'))
+
+    else:
+        work_list = list(itertools.product(variables, plot_types))
```
CoryMartin-NOAA: where is `work_list` used?

LAMDA/scripts/LAMDA_stats_workflow.py
```diff
@@ -0,0 +1,122 @@
...
+
+    # Create multiprocessing Pool
+    p = Pool(processes=nprocs)
+    p.map(partial(plotting_func, data_dict=data_dict,
+                  outdir=outdir), work)
```
CoryMartin-NOAA: should this be `work_list`?

CoryMartin-NOAA (commented):
Overall good, but a few comments/questions.

LAMDA/map_qc_flags.py
kevindougherty-noaa: Right I got some information on that a ways back. Is there one that we prefer or do we need both? Is this just for temperature alone?

LAMDA/plot_features.py
kevindougherty-noaa: For our multiprocessing workflow, just one. But it has the capability to include multiple if someone wanted to use the individual scripts.

LAMDA/scripts/LAMDA_stats_workflow.py
kevindougherty-noaa: Good catch.

LAMDA/scripts/LAMDA_stats_workflow.py
kevindougherty-noaa: Yes, it should be work_list, not work.

LAMDA/map_qc_flags.py
CoryMartin-NOAA: We need both, for temperature only the `setup_qc_mark` will just tell you if it is Tv or T. The `prep_qc_mark` is actually the QC flag from the prepBUFR preprocessing step.

LAMDA/scripts/LAMDA_mapping_workflow.py
```diff
@@ -0,0 +1,107 @@
+import yaml
```
CoryMartin-NOAA, line 1: is this script intended to be 'executable'? if so, add a shebang at the top?
`#!/usr/bin/env python3`

LAMDA/scripts/LAMDA_stats_workflow.py
```diff
@@ -0,0 +1,189 @@
...
+import os
+import pandas as pd
+import numpy as np
+import itertools
+from multiprocessing import Pool
+from functools import partial
+from datetime import datetime
+from pyGSI.gsi_stat import GSIstat
+
+# import the plotting scripts i.e:
```
CoryMartin-NOAA, line 12: why are these commented out?

LAMDA/scripts/LAMDA_stats_workflow.py
```diff
@@ -0,0 +1,189 @@
+import yaml
```
CoryMartin-NOAA, line 1: Same comment as previous file, are these intended to be executable scripts?

LAMDA/scripts/LAMDA_stats_workflow.py
```diff
@@ -0,0 +1,189 @@
...
+    elif data_type == 'radiance':
+        # change ob_type to just the sensor to utilize
+        # GSIstat extract_sensor()
+        ob_type = ob_type.split('_')[0]
+        plotting_config['channel'] = data_inputs[0]
+        plotting_config['obuse'] = data_inputs[-1]
+
+    # Loop through t-minus hours to grab proper data and
+    # generate plots
+    for tm in np.arange(7):
```
CoryMartin-NOAA, line 84: For now I think this is fine, but I think eventually we might need to check to make sure the tm06-tm01 files exist as some cycles might just be tm00 for operations?

LAMDA/scripts/LAMDA_stats_workflow.py
```diff
@@ -0,0 +1,189 @@
...
+    functions purpose is to generate the minimization plots.
+
+    Args:
+        data_dir : (dict) dictionary that includes fits2 data
+        outdir : (str) path to where figures should be outputted
+    """
+
+    # Loop through t-minus hours to grab proper data and
+    # generate plots
+    for tm in np.arange(7):
```
CoryMartin-NOAA, line 114: Same comment.

CoryMartin-NOAA (commented):
Main comment/question is are the scripts with `argparse` intended to run like so:

`script.py --arg1 foo --arg2 bar`?

If so, we need `#!/usr/bin/env python3` at the top.
Otherwise, they need to be called like so:
`python script.py --arg1 foo --arg2 bar` which is fine, just checking design intentions.

LAMDA/scripts/LAMDA_stats_workflow.py
kevindougherty-noaa: They do not exist currently. This is just an example for the moment how they would be added when plotting scripts are created and imported.

LAMDA/scripts/LAMDA_stats_workflow.py
kevindougherty-noaa: Got it. What I can probably do is get the length of the data dict list for the fits and fits2 data.

LAMDA/scripts/LAMDA_mapping_workflow.py
kevindougherty-noaa: Whatever you prefer is fine. They will be called from a shell script in the parallelization so it's up to if you want to include the python command or not.

LAMDA/scripts/LAMDA_stats_workflow.py
CoryMartin-NOAA: Let's leave that for now I think and merge this in so that we are prepared for next week.

CoryMartin-NOAA (approved):

PR #91, opened 2021-11-10
Allow for longer line length

CoryMartin-NOAA:
in the pycodestyle CI tests

### DIFF
--- a/.pycodestyle
+++ b/.pycodestyle
@@ -1,0 +1,2 @@
+[pycodestyle]
+max-line-length = 160


aerorahul (approved):
You
Might want to add some other commonly ignores errors. 
Like E401 that ignores multiple imports on the same line. 

E.g. 
`import os, sys` would be an error if 401 is not ignored. 
Your call though.

CoryMartin-NOAA:
@aerorahul I'm actually okay with keeping that style convention, because then it's hard to do `import foo as bar` or `from foo import bar`

PR #108, opened 2021-11-17
Create Basic OmF/OmA Maps

### LINKED ISSUE #96: Spatial Plots of O-F/O-A for LAMDA Diagnostics
As part of the LAMDA monitoring code sprint, create figures of O-F and O-A values for various instruments/obs types that plot on multiple regional domain maps.

CoryMartin-NOAA:
Closes #96 

This adds a new type of plot, `map departures`, which will plot O-F or O-A (depending on YAML) for the specified domains and variables.

It computes the vmin/vmax based off of the max/min values and is always centered on 0, and also annotates the nobs, min, max on the figure.

Two examples are attached.
Screen Shot 2021-11-17 at 3 54 37 PM
Screen Shot 2021-11-17 at 3 57 14 PM

### DIFF
--- a/LAMDA/map_departures.py
+++ b/LAMDA/map_departures.py
@@ -37,7 +37,7 @@ def _create_map_departures(df, domain, projection,
         omfscatter = MapScatter(latitude=lats,
                                 longitude=lons,
                                 data=omf)
-        omfscatter.markersize = 1
+        omfscatter.markersize = 2
         omfscatter.cmap = 'coolwarm'
         # determine min/max of values
         maxval = np.nanmax(omf)
@@ -62,8 +62,8 @@ def _create_map_departures(df, domain, projection,
         # add stats to figure
         stats_dict = {
             'nobs': len(lats),
-            'min': minval,
-            'max': maxval,
+            'min': np.round(minval, 4),
+            'max': np.round(maxval, 4),
         }
         mymap.add_stats_dict(stats_dict=stats_dict)
         # save figure
--- a/LAMDA/map_departures.py
+++ b/LAMDA/map_departures.py
@@ -62,8 +62,8 @@ def _create_map_departures(df, domain, projection,
         # add stats to figure
         stats_dict = {
             'nobs': len(lats),
-            'min': np.round(minval, 4),
-            'max': np.round(maxval, 4),
+            'min': str(np.round(minval, 4)),
+            'max': str(np.round(maxval, 4)),
         }
         mymap.add_stats_dict(stats_dict=stats_dict)
         # save figure
--- a/LAMDA/map_departures.py
+++ b/LAMDA/map_departures.py
@@ -89,7 +89,7 @@ def map_departures(config):
                            analysis_use=config['analysis use'])
         metadata = diag.metadata
 
-    metadata['Diag Type'] = 'Departures'
+    metadata['Diag Type'] = config['data type']
     metadata['Anl Use Type'] = None
 
     # pass dataframe to plot function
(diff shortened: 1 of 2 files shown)


LAMDA/map_departures.py
```diff
@@ -0,0 +1,120 @@
...
+        mymap.add_title(labels['title'], loc='left', fontsize=12)
+        mymap.add_title(labels['date title'], loc='right', fontsize=12,
+                        fontweight='semibold')
+        mymap.add_xlabel("Longitude", fontsize=12)
+        mymap.add_ylabel("Latitude", fontsize=12)
+        # add stats to figure
+        stats_dict = {
+            'nobs': len(lats),
+            'min': minval,
+            'max': maxval,
```
kevindougherty-noaa: Can you use `np.round()` to maybe 3 or 4 decimal places?

kevindougherty-noaa (changes requested):
Looks great, just one minor comment.

JacobCarley-NOAA (changes requested):
Maps look good. Some marks are hard to see when values are small. Suggest adding nonzero line thickness to the marker (circle) edges to make the locations of each observation more clear.

LAMDA/map_departures.py
CoryMartin-NOAA: Good idea, will do so in the next commit.

LAMDA/map_departures.py
```diff
@@ -37,7 +37,7 @@ def _create_map_departures(df, domain, projection,
         omfscatter = MapScatter(latitude=lats,
                                 longitude=lons,
                                 data=omf)
-        omfscatter.markersize = 1
+        omfscatter.markersize = 2
```
JacobCarley-NOAA, line 40: Possible to add the following kwarg? `omfscatter.linewidths=0.15`. I don't know API from which MapScatter is derived, but assuming it's from matplotlib and is based upon `matplotlib.pyplot.scatter` that should help the markers show up a little bit better. If it doesn't work right we made need to change the edgecolor kwarg.

CoryMartin-NOAA: I think this is something we will have to add to EMCPy, as I don't think all **kwargs are inherited. We probably have to add it, what do you think @kevindougherty-noaa ?

kevindougherty-noaa: Yes, I was just about to comment on this. Taking a look at [MapScatter](https://github.com/NOAA-EMC/emcpy/blob/develop/src/emcpy/plots/map_plots.py#L1), I think it would be just as simple as adding `self.linewidths` as an additional argument.

CoryMartin-NOAA: I think this needs even more work in EMCPy, as I don't think the edge colors argument is passed properly, see https://github.com/NOAA-EMC/emcpy/issues/80

CoryMartin-NOAA:
@JacobCarley-NOAA thoughts? I think this now makes the line edges dominate the plot instead of the fill colors:
Screen Shot 2021-11-18 at 9 18 05 AM
Screen Shot 2021-11-18 at 9 19 37 AM

kevindougherty-noaa:
@CoryMartin-NOAA I agree it kind of takes away from the colors. Also, this is not an issue on your end, but it looks like CONUS domain is using a 110m coastline and 50m states. Will have to figure that out in EMCPy how to handle that better.

JacobCarley-NOAA:
@CoryMartin-NOAA Yeah. You and @kevindougherty-noaa are right. I'm kind of surprised of the overwhelming effect of adding the edges to be honest. No edges is definitely preferable over this. 

Overall this is pretty minor.

JacobCarley-NOAA (approved):

CoryMartin-NOAA:
@kevindougherty-noaa do you approve the PR now? You requested changes earlier.

kevindougherty-noaa (approved):
Yes, my apologies. Looks great!

PR #112, opened 2021-11-22
Add bias and std. dev. per channel diagnostic

### LINKED ISSUE #111: Create diagnostic to plot OmF and Standard Deviation averages per channel
Create a diagnostic similar to the current monitoring page diagnostic that displays bias and standard deviation per channel for a single cycle and average over multiple cycles:
image

kevindougherty-noaa:
Closes #111 

I attempted to replicate the current diagnostic that exists for global monitoring that plots the bias (with bias correction) and std deviation per channel. This has only been tested for non hyperspectral models and will need to be updated on how to handle only subsets of channels. 

Example plot shown here:
image


JacobCarley-NOAA (changes requested):
Suggest using the same color but different linestyle for the 5 day averages as their counterpart stat. For example, O-F is blue and the 5day average is also blue but is a bold dashed line (or something like that).

kevindougherty-noaa:
Updated for @JacobCarley-NOAA's suggestion.
image

JacobCarley-NOAA (approved):
looks good!

PR #115, opened 2022-01-28
Adding LAMDA filtering toolkit

### LINKED ISSUE #114: Add toolkit to help with filtering data
To start, there will be options to filter land, water, clouds, snow, ice, and vegetation.

kevindougherty-noaa:
Closes #114


LAMDA/LAMDA_tools.py
```diff
@@ -0,0 +1,56 @@
+import numpy as np
+import pandas as pd
+
+
+class LAMDA_Tools:
+
+    def __init__(self):
```
CoryMartin-NOAA: Should this be a class?
So you would call it like so:
```
tool = LAMDA_Tools()
land_df = tool.land_fraction(all_df)
```

Shouldn't these just be functions? Or am I missing something.

kevindougherty-noaa: Yes, you are right.

CoryMartin-NOAA (approved):

PR #118, opened 2022-02-22
bug fix to pyGSI/gsi_stat.py ; added aggregate rms/bias option to scripts/plot_gsi_stat_exp.py

BrettHoover-NOAA:
**pyGSI/gsi_stat.py**:
L266: the search/replace string for header has to be a FORTRAN-formatted expression, rather than a simple float. So '2000.0' needs to be replaced with '0.200E' + r'\+04', in order for the header to read 'column' for this entry.

**scripts/plot_gsi_stat_exp.py**: 
L13: added variable plottype to allow for choosing either the mean rms/bias ('mean') or aggregate rms/bias ('aggr') in gen_figure()
L31-32: removed the implicit declaration for stattype, now stattype is user-defined in call to gen_figure() via plottype
L182,184: added aggregate rms and bias scores to rmses and biases dictionaries
L194-212: aggregate rms/bias scores are computed, being careful to assign np.nan to any level with counts=0
L216-218: gen_figure() is passed plottype variable to plot either mean or aggregate rms/bias scores

### DIFF
--- a/pyGSI/gsi_stat.py
+++ b/pyGSI/gsi_stat.py
@@ -263,7 +263,7 @@ class GSIstat(object):
             if re.search(pattern, line):
                 header = line.strip()
                 header = re.sub('pbot', 'stat', header)
-                header = re.sub('2000.0', 'column', header)
+                header = re.sub('0.200E\+04', 'column', header)
                 pbots = np.array(header.split()[7:-1], dtype=np.float)
                 break
         if pbots is []:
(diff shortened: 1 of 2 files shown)


kevindougherty-noaa:
@BrettHoover-NOAA Looks good to me. You just need to request a review from myself so I can approve and merge.

kevindougherty-noaa (approved):
Never mind, I can request myself.

PR #123, opened 2022-03-22
Plot subtype

BrettHoover-NOAA:
Includes scripts/plot_proftrace.py, which produces two figures for chosen types/subtypes of conventional data:

1) (left) vertical profiles of aggregate rmse and bias, (right) bar-plot of obs-counts per pressure level
2) (top) time-series trace of aggregate rmse and bias, (bottom) bar-plot of obs-counts per analysis period

Can accommodate up to 10 different subsets of conventional data, each of which is selected by adding a dictionary of the following form to the 'set_dict' list:

```
{ 
'name' : <input option>, 
'var' : <input option>, 
'it' : <filter option>, 
'use' : <filter option>, 
'typ' : <filter option>, 
'styp' :<filter option>,
'statdir' : <input option>
}
```
'name' is a string that is used to identify the data (sub)set in figure legends
'var' is a string identifying the type of conventional data being plotted ('uv', 't', or 'q')
'it' is an integer filter option specifying which iteration-loops the O-F is computed from. Options are 1, 2, or 3
'use' is a string filter option specifying which use-types are included. Options are 'asm', 'mon', or 'rej'
'typ' is an integer filter option specifying which ob-types are included.
'styp' is a string filter option specifying which ob-subtypes are included.
'statdir' is a string used to identify the full-path to the gdas.gsistat.* files

Filter options can be provided as any of the following:
* a single value, which excludes all data not defined by the filter option (e.g., 'typ': 220 would exclude everything except rawinsonde wind observations)
* a list of values, which excludes all data not meeting at least one filter option (e.g., 'use': ['asm','mon'] would include any assimilated or monitored observation, but would exclude rejected observations)
* None, which includes any data (e.g., 'styp':None would include any subtype of the data)

For example, to include OMB from any assimilated rawinsonde temperature observation:
```
set_dict.append({ 
'name' : 'RAOB-t', 
'var' : 't', 
'it' : 1, 
'use' : 'asm', 
'typ' : 120, 
'styp' : None,
'statdir' : </full/path/to/gsistat/files/>
})
```

To include OMA from monitored GOES-15 short-wave IR winds:
```
set_dict.append({ 
'name' : 'G15-SWIR-winds', 
'var' : 'uv', 
'it' : 3, 
'use' : 'mon', 
'typ' : 240, 
'styp' : 259,
'statdir' : </full/path/to/gsistat/files/>
})
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -82,7 +82,8 @@ profile_default/
 ipython_config.py
 
 # Figure files
-.png
+*.png
+
 # pyenv
 .python-version
 
--- a/.gitignore
+++ b/.gitignore
@@ -81,6 +81,8 @@ target/
 profile_default/
 ipython_config.py
 
+# Figure files
+.png
 # pyenv
 .python-version
 
--- a/plot_proftrace.py
+++ b/plot_proftrace.py
@@ -1,0 +1,395 @@
+import numpy as np
+import matplotlib.pyplot as plt
+import pandas as pd
+from pyGSI.gsi_stat import GSIstat
+import os
+
+def get_datelist(startdate, enddate):
+    import datetime
+    datelist = []
+    mydate = startdate
+    while mydate <= enddate:
+        date = datetime.datetime.strptime(mydate,'%Y%m%d%H')
+        datelist.append(mydate)
+        nextdate = date + datetime.timedelta(hours=6)
+        mydate = datetime.datetime.strftime(nextdate,'%Y%m%d%H')
+    return datelist
+
+########################
+#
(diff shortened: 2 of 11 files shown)


kevindougherty-noaa:
@BrettHoover-NOAA Have you considered using a yaml input so users only have to edit the yaml and not the source code for your dictionary inputs?

BrettHoover-NOAA:
Hi Kevin,

I'm not very familiar with yaml inputs, but I could definitely try going
that rout if it's a less klugey way to navigate options. I can close the
current pull request and give that a try

PR #124, opened 2022-03-23
Plot subtype

BrettHoover-NOAA:
A script that produces two figures for conventional data from gsistat.gdas.* files:
1) A vertical profile of aggregate rmse and bias (left), with observation counts per pressure bin (right)
2) A time-series trace of full-column rmse and bias (top), with observation counts per analysis time (bottom)

Plotting options are controlled by the proftrace_yaml.yaml file. This file contains a section for figure options:
```
figures:
    pname: <name of profile figure-file>
    tname: <name of trace figure-file>
    tskip: <number of ticks to skip when plotting tick-labels, to prevent labels from overwriting each other>
```
There is also a section for defining dates over which gsistat files are collected:
```
dates:
    dt_beg: <date of beginning of time-series>
    dt_end: <date of end of time-series>
    hrs: <hours between each time period between dt_beg and dt_end>
```
Each remaining section of the file specifies a set of filters to generate scores for a data (sub)set:
```
set:
    name: <name of data (sub)set on plot legends>
    var: <conventional variable-type>
    it: <FILTER: loop-iteration of DA to pull statistics from>
    use: <FILTER: use-flag of data>
    typ: <FILTER: data type>
    styp: <FILTER: data subtype>
    statdir: <full path to directory containing gsistat files>
```

Each filter option (it, use, typ, styp) can be one of three types:
1) A single value, producing a subset that corresponds to only that value. For example, to subset only it=1 (unadjusted background):
```
    it: 1
```
2) A list of values, producing a subset of observations matching any option in the list. For example, to subset both assimilated and rejected use-types:
```
    use: ['asm', 'rej']
```
3) A value of None, producing a subset of observations that bypasses the filter and accepts all types. For example, to subset including all subtypes:
```
    styp: None
```

Script is executed in the PyGSI environment:
```
python plot_proftrace.py
```

### DIFF
--- a/scripts/plot_proftrace.py
+++ b/scripts/plot_proftrace.py
@@ -1,10 +1,61 @@
-import numpy as np
-import matplotlib.pyplot as plt
-import pandas as pd
-from pyGSI.gsi_stat import GSIstat
-import os
-
-
+##########################################################################
+#
+# YAML FILE CONTENTS
+#
+# yaml file contents come in three types, based on the key name:
+#
+# figures: Defines settings for profile and trace figures
+#    pname: Name of profile figure-file
+#    tname: Name of trace figure-file
+#    tskip: Number of ticks that are skipped when plotting tick-labels on
+#           trace-figure (should be large enough to keep labels from
+#           stepping on each other)
+#
+# dates: Defines settings for time-period over which statistics are
+#        collected and plotted
+#    dt_beg: Beginning date of time-series (%Y%m%d%H format)
+#    dt_end: Ending date of time-series (%Y%m%d%H format)
+#    hrs: Number of hours between each time-period between dt_beg and
+#         dt_end (usually 6 for collecting every cycle)
+#
+# <any other key name>: Each member of the plot is provided as a separate
+# dictionary with the following keys:
+#    name: Name of member, used in the legend of the plot to distinguish
+#          lines or bars belonging to this set
+#    var: Variable being extracted ('uv', 't', or 'q')
+#    it: Iteration(s) (1, 2, or 3), can be an individual value or a list
+#        of values (e.g. [1,3]), or None to collect all values
+#    use: Use-type ('asm', 'mon', or 'rej'), can be an individual value
+#         or a list of values (e.g. ['asm','rej']), or None to collect all
+#         values
+#    typ: Observation type number, can be an individual value or a list of
+#         values, or None to collect all values
+#    styp: Observation subtype string, can be an individual value or a
+#          list of values, or None to collect all values
+#    statdir: String identifying full-path to directory containing gsistat
+#             files (ends in '/')
+##########################################################################
+# The list of cycles to extract data from, which will be aggregated for
+# profiles and plotted individually for traces, is defined by:
+#    1) Use of get_datelist() to automatically generate a list of cycles
+#       at regular steps between a beginning and ending date, e.g.:
+#
+#       cycles=get_datelist(dt_beg,dt_end,hrs)
+#
+#    2) During search for gsistat files for each cycle, if a gsistat file
+#       is missing, the date is removed from cycles
+#    3) During collection of statistics, if there are no valid statistics
+#       for a given cycle, NaN values are returned and plotted
+#
+# Figures are saved to pname and tname, respectively
+#
+# tskip defines the number of time-periods in the trace that are skipped
+# when adding tick-labels. Tick-labels are dates of the form (e.g.)
+# 'Aug01-00z', so for long timeseries traces the labels will be written
+# over each other unless some of them are skipped. I usually use some
+# multiple of 4, so that the tick-labels always correspond to the same
+# analysis-period.
+##########################################################################
 def get_datelist(startdate, enddate, hrs):
     import datetime
     datelist = []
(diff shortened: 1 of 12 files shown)


scripts/plot_proftrace.py
```diff
@@ -0,0 +1,622 @@
...
+                tskip = parsed_yaml['figures']['tskip']
+            except MissingFiguresError:
+                pname = None
+                tname = None
+                tskip = None
+                print('Error reading key: figures')
+            # Extract all other keys as filtering dictionaries, store
+            # in setdict list
+            setdict = []
+            for key in parsed_yaml.keys():
```
kevindougherty-noaa: I think you can simplify up this part a bit:

```
filt_dict = {x: parsed_yaml[x] for x in parsed_yaml if x not in {'dates', 'figures'}}

filt_keys = ['it', 'use', 'typ', 'styp']
for key in filt_keys:
    try:
        filt_dict[key] = None if filt_dict[key] == 'None' else filt_dict[key]
        setdict.append(filt_dict)
    except FilterDictReadError:
        print('Error reading filter-dictionary key:', key)
```

scripts/plot_proftrace.py
```diff
@@ -0,0 +1,622 @@
...
+    # Set legend
+    plt.legend(nameList, frameon=False, fontsize=10)
+    # Set y-label
+    plt.ylabel('Ob Count (Thousands)')
+    # Turn off interactive-mode to suppress plotting figure
+    plt.ioff()
+    # Return
+    return plt.gcf()
+
+
```
kevindougherty-noaa, line 652: Just good Python practice to include this line:
```
if __name__ == "__main__":
```
And then tab everything over below this line.

scripts/plot_proftrace.py
```diff
@@ -0,0 +1,622 @@
...
+                                              slice(None), slice(None),
+                                              'rms'), s.columns[-1]]
+        counts[setname][cycle]['col'] = s.loc[(slice(None), slice(None),
+                                               slice(None), slice(None),
+                                               slice(None), slice(None),
+                                               'count'), s.columns[-1]]
+        biases[setname][cycle]['col'] = s.loc[(slice(None), slice(None),
+                                               slice(None), slice(None),
+                                               slice(None), slice(None),
+                                               'bias'), s.columns[-1]]
```
kevindougherty-noaa: I could possibly see this whole section of code as a function that creates the dictionaries based on specific inputs to save lines of code.

scripts/plot_proftrace.py
```diff
@@ -0,0 +1,622 @@
...
+            cycles.remove(cycle)
+            raise FileNotFoundError(f'Unable to find {inputfile} for {cycle}')
+        # statistics are drawn for a single variable
+        var = sd['var']
+        # all other filters (it,use,typ,styp) could contain multiple
+        # entries in a list
+        # if they are not a list, assert them as a list
+        it = sd['it']
+        if type(it) != list:
+            it = [it]
```
kevindougherty-noaa:
```
it = [it] if type(it) != list else it
```

scripts/plot_proftrace.py
```diff
@@ -0,0 +1,622 @@
...
+        use = sd['use']
+        if type(use) != list:
+            use = [use]
```
kevindougherty-noaa: Same as above

scripts/plot_proftrace.py
```diff
@@ -0,0 +1,622 @@
...
+        typ = sd['typ']
+        if type(typ) != list:
+            typ = [typ]
```
kevindougherty-noaa: Same as above

scripts/plot_proftrace.py
```diff
@@ -0,0 +1,622 @@
...
+        styp = sd['styp']
+        if type(styp) != list:
+            styp = [styp]
```
kevindougherty-noaa: Same as above

scripts/plot_proftrace.py
```diff
@@ -0,0 +1,622 @@
...
+    rmse[nobs == 0] = np.nan
+    bias[nobs == 0] = np.nan
+    r_trace[n_trace == 0] = np.nan
+    b_trace[n_trace == 0] = np.nan
+    rmse_trace.append(r_trace)
+    bias_trace.append(b_trace)
+    nobs_trace.append(n_trace)
+    rmse_profs.append(rmse)
+    bias_profs.append(bias)
+    nobs_profs.append(nobs)
```
kevindougherty-noaa: Could this also be a function that can be called for each dictionary? Not a big deal just trying to minimize lines of code.

kevindougherty-noaa (commented):
Just a few suggestions where code can be more pythonic and decrease the number of lines of code. Thank you for the detailed documentation.

BrettHoover-NOAA:
Hi @kevindougherty-noaa, I appreciate all of the advice on tightening up the language (the pythonics?) in the code. I will take a crack at making these suggested changes and get back to you.

BrettHoover-NOAA:
@kevindougherty-noaa: I took a crack at some of the suggestions you had, and rearranged the code with more discrete sets of functions on top and a main section at the bottom that just sequences the functions. There wasn't a way to generalize the collect_statistics and aggregate_figure_data functions that wouldn't have either been clunky or increased code size, but the execution should (hopefully!) be clearer. I also had to fix the exception handling because I saw errors, and went through and tested all of the exceptions to make sure they are working now. Let me know what you think, thanks!

kevindougherty-noaa (approved):
Looks good to me. Thanks for making those changes!

PR #131, opened 2022-12-22
Overhaul to plot_proftrace: options for multiple figure configurations in YAML, plotting per-ob DA penalty statistics

BrettHoover-NOAA:
This is primarily a change to plot_proftrace.py and its associated YAML file proftrace_yaml.yaml, but there is a critical change to pyGSI/gsi_stat.py as well.

DA statistics plotting in plot_proftrace.py is expanded to include the following:

- Instead of defining a single set of filtering and plotting options to produce profile and trace figures for a single dataset, the YAML is now configured into FIGURE CARDS, each of which can contain its own filtering/plotting options to produce profile and/or trace figures for a different dataset. In effect, rather than repeatedly changing the YAML to produce figures over different datasets, you can now append all of these plotting decisions into individual sets (FIGURE CARDS), and plot_proftrace.py will iterate through each FIGURE CARD to produce figures. This is useful if, for example, you need to generate a lot of similar plots across different observation types, in which case you can construct one YAML file for your experiment to generate all of the figures.
- Options are now added to produce either DA error (rms and bias) profiles and traces, or DA penalty (per-ob penalty: `cpen`) profiles and traces. It can be useful to look at differences in DA penalties between an experiment and control when, for example, you are testing a DA change that does not appreciably change the observations being assimilated but instead how those observations are assimilated. Turning off observation downweighting for duplicate observations, for example, would have no direct impact on the number or type of observations being assimilated, but could have impacts on the penalties assigned to those observations.
- You now have the option to generate any of the four figure types (error profile, error trace, penalty profile, or penalty trace) individually in a FIGURE CARD, through choice of the figure-name YAML keys `figopts: errProName`, `figopts: errTraName`, `figopts: penProName`, `figopts: penTraName`, respectively. Setting any of these figure-name keys to `None` will skip drawing that figure, and likewise not assigning the figure-name key at all will skip drawing that figure.
- In order to allow for plotting DA penalty statistics, pyGSI/gsi_stat.py was changed in `_get_conv` to retain the `cpen` and `qcpen` data, where previously these data were being excluded.

Example YAML has been updated to FIGURE CARDS with varying options to generate different figures.


proftrace_yaml.yaml
```diff
@@ -1,24 +1,55 @@
...
-     hrs: 6
+LWIR_figs:
+     set1:
+          name:    'LWIR-G17'
+          var:     'uv'
+          it:      1
+          use:     'asm'
+          typ:     245
+          styp:    '0271'
+          statdir: /Users/bhoover/Desktop/IMSG/PROJECTS/GOES18/impact_test/stat_tank/v163ctl_sept/
```
kevindougherty-noaa: My only worry is that this and others below are a local directory. Are these just stat files? Do you think it would be beneficial to add to the repository? Maybe added to the `PyGSI/fix` directory if they are just text files?

kevindougherty-noaa (commented):
Thank you for your patience and my apologies as I just saw this PR this morning. I don't see anything major, I just had one comment about the statdirs in `proftrace_yaml.yaml` before I approve.

kevindougherty-noaa (approved):
Looks good to me.

proftrace_yaml.yaml
BrettHoover-NOAA: Hi @kevindougherty-noaa – You're right, these were accidentally left in as pointing to local directories containing the gsistat files I was using during testing. I replaced them with generic `/path/to/set#/gsistat/files` paths to indicate that a user should change these to point to their respective repository of stat files.

The rest of the settings were also test settings, but I think it would be useful to leave in real examples here so people can see what the format is (some are strings, some are integers), and recognize what the options are referencing.

PR #133, opened 2023-02-03
Feature/obspace diagnostics

delippi:
Added some capabilities for observation space ensemble DA diagnostics for RRFS. Two main scripts were added: filter_obs.py and obspace.py. The filter obs script is a function for selecting which observations to use from the diag file. This can filter observations by observation type from the bufr obs table (https://emc.ncep.noaa.gov/mmb/data_processing/prepbufr.doc/table_2.htm) as well as by lat/lon/pressure/error. The obspace.py script computes and plots the observation space ensemble DA diagnostics.

### DIFF
--- a/pyGSI/obspace.py
+++ b/pyGSI/obspace.py
@@ -55,7 +55,7 @@ def ensemble_obspace_diag(
       std_dev       : (array float) standard deviation of (F-O)
       spread        : (array float) ensemble spread (standard deviation)
       ob_error      : (array float) observation error standard deviation
-      total_spread  : (array float) total spread (standard deviation) 
+      total_spread  : (array float) total spread (standard deviation)
       num_obs_total : (array float) total number of observations
       num_obs_assim : (array float) total number of observations assimilated
       cr            : (array float) consistency ratio (total spread/rmsi)**2
@@ -189,10 +189,11 @@ def ensemble_obspace_diag(
                     mem = mem + 1
                     # end while n_mem
 
-                try: # if we broke out of the loop above before, because we were missing a mem.
-                     # this will break out of the loop here too.
+                # if we broke out of the loop above before, because we were missing a mem.
+                # this will break out of the loop here too.
+                try:
                     error_var = error**2
-                except:
+                except ValueError:
                     break
 
                 fcst_ens_mean = fcst_ens_mean / n_mem
@@ -237,8 +238,6 @@ def ensemble_obspace_diag(
                 del lon
                 del pressure
                 del use
-                #del 
-                #del 
 
             # end do n_expt
         # end do n_ob_type
--- a/pyGSI/obspace.py
+++ b/pyGSI/obspace.py
@@ -1,0 +1,230 @@
+import numpy as _np
+import os as _os
+import emcpy.utils.dateutils as _dateutils
+import emcpy.io as _io
+import pyGSI.filter_obs as _filter_obs
+
+
+def ensemble_obspace_diag(
+    datapath,
+    date1,
+    date2,
+    expt_names,
+    n_mem,
+    skip_enkf_hours=[],
+    ob_types=["u"],
+    codes_uv=[187],
+    codes_tq=[287],
+    hem=None,
+    p_max=1050.0,
+    p_min=100.0,
+    lat_max=90.0,
+    lat_min=0.0,
+    lon_max=360.0,
+    lon_min=0.0,
+    error_max=40.0,
+    error_min=0.000001,
+):
+    """
+    Computes observation space stats into 3D arrays (n_ob_type, n_expt, 24) and plots the stats as a
+    function of hour of day.
+
+    Args:
+      datapath        : (str) netCDF filename
+      date1           : (str "YYYYMMDDHH") start date
+      date2           : (str "YYYYMMDDHH") end date
+      skip_enkf_hours : (list of int) hours (UTC) to skip for enkf
+      expt_names      : (list of str) experiment names
+      n_mem           : (int) number of ensemble members
+      ob_types        : (list of str) observation types (u,v,t,q,etc.)
+      codes_uv        : (list of int) uv bufr codes used to filter obs
+      codes_tq        : (list of int) tq bufr codes used to filter obs
+      p_max           : (float) maximum pressure (mb) for including observation in calculations
+      p_min           : (float) minimum pressure (mb) for including observation in calculations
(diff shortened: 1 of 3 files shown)


pyGSI/obspace.py
```diff
@@ -0,0 +1,245 @@
...
+    for date in dates:
+        times = _dateutils.datetohrs(date)
+        pdy = str(date[0:8])
+        hour = int(str(date[8:10]))
+        if any(item == hour for item in skip_enkf_hours):
+            continue  # As of right now, the rrfs only runs the EnKF at 18-23Z so skip those off times
+        for ob_type in ob_types:
+            i_o = ob_types.index(ob_type)
+            for expt_name in expt_names:
+                print("%s %s %s" % (date, expt_name, ob_type))
```
CoryMartin-NOAA: I suggest using "f-strings" here and below in the `_os.path.join`. I'll show an example there.

pyGSI/obspace.py
```diff
@@ -0,0 +1,245 @@
...
+                i_e = expt_names.index(expt_name)
+                mem = 1
+                while mem <= n_mem:
+                    memid = str(mem).zfill(4)
+                    if ob_type == "u" or ob_type == "v":
+                        obsfile = _os.path.join(datapath, "%s/%s/mem%s/diag_conv_uv_ges.%s.nc4" % (expt_name, date, memid, date))
```
CoryMartin-NOAA:
```suggestion
                        obsfile = _os.path.join(datapath, f"{expt_name}/{date}/mem{memid}/diag_conv_uv_ges.{date}.nc4")
```

pyGSI/obspace.py
```diff
@@ -0,0 +1,245 @@
...
+                    else:
+                        obsfile = _os.path.join(datapath, "%s/%s/mem%s/diag_conv_%s_ges.%s.nc4" % (expt_name, date, memid, ob_type, date))
+
+                    exists = _os.path.exists(obsfile)
+                    if not exists:
+                        print("diag file for %s mem%s %s doesn't exist." % (expt_name, str(mem).zfill(4), str(date)))
+                        mem = mem + 1
+                        break
```
CoryMartin-NOAA: is this okay?

pyGSI/obspace.py
```diff
@@ -0,0 +1,245 @@
...
+
+                    if mem == 1:
+                        code = _io.netCDF.read_netCDF_var(obsfile, "Observation_Type", oneD=True)
```
CoryMartin-NOAA: Seems like it would be cleaner to use the existing PyGSI interface for reading netCDF diag files?

CoryMartin-NOAA (commented):
A few small comments/questions.

pyGSI/obspace.py
delippi: I'm not sure if I see how to implement that. Could you possibly provide an example @CoryMartin-NOAA ? What I have, at least to me, seems to already be fairly clean and easy to understand... I thought.

CoryMartin-NOAA: Have you looked at the https://github.com/NOAA-EMC/PyGSI/blob/39693c00d879f6fa30d7664808b60bd604501255/pyGSI/diags.py#L41 conventional class? It may end up being not worth using this, just a suggestion because Kevin spent a lot of time putting diag contents into classes/data frames

delippi: @CoryMartin-NOAA , okay I see, thank you. I think it would be ideal to use this, if possible. I made a version of the code which uses this, but I am currently unable to get it to work as efficiently. I think it is in part due to WCOSS being a bit laggy for me today. I will check again on Monday, but I still don't think it will be as quick. There is also the possibility that I made an error somewhere. I will double check this with fresh eyes on Monday as well.

CoryMartin-NOAA: Sounds good, thanks for trying to use it. I think it makes the total code base much smaller by not duplicating read routines.

delippi: @CoryMartin-NOAA, agreed. If it can be used it should, but I'm still having a little trouble. Here is my problem. The same observations are used by each ensemble member (in terms of what is fed into them), right? But, as you know, each ensemble member may actually assimilate a different number of observations (e.g., possibly due to gross checks etc.). I'm finding it a little difficult to only consider the observations assimilated by all ensemble members under the current read routines (using pandas data frames); it could be just that I haven't figured out a good solution as I'm sure it can be done. This also highlighted an issue with how I was implementing my routines which has since been fixed to only consider the observations assimilated by all members. If this is a major sticking point, I'm happy to discuss further details in a meeting and maybe we can do a small code sprint to figure it out.

CoryMartin-NOAA: So is the issue combining data frames and matching their indices? Or is it something else? I would prefer it to be using the existing IO routines but if that is too much effort, we can merge in your specific use case for now, note it, and refactor later if necessary?

delippi: @CoryMartin-NOAA, yes, the problem is combining data frames and matching indices. I'll send you and email to work out a possible solution.

delippi: @CoryMartin-NOAA , this was the solution we discussed yesterday:
```
from pyGSI.diags import Conventional as _Conventional
diag = _Conventional(diagfile)
data = diag.get_data(obsid=codes, analysis_use=False)
```
Using `analysis_use=False` makes the problem more simple, but it doesn't speed it up any at all. It still requires a lot of overhead reading in the entire dataset into a pandas data frame (i.e., `diag = _Conventional(diagfile)`) for each ensemble diag file.

Here are the timings for 1 cycle time, 4 variables (u, v, t, q), and a 30 member ensemble:
```
Method using pandas data frames (existing method in PyGSI): Total elapsed time: (209.82s), 00:03:29
Method that I plan to use (use netCDF4 Dataset to read data): Total elapsed time: (1.66s), 00:00:01
```
That comes out to be about 1-2s to read each diag file which isn't bad if you're only reading one.

CoryMartin-NOAA: ah, okay, then yep, proceed with your current approach, pandas is too slow for this

pyGSI/obspace.py
delippi: @CoryMartin-NOAA , I've since changed the name of this script, but it is needed in the case there might be missing data over the data range that I run the scripts.

CoryMartin-NOAA (approved):

PR #134, opened 2023-02-23
Clean up repository

kevindougherty-noaa:
Removing outdated scripts, yamls, and functions.


kevindougherty-noaa:
All scripts have been checked and are working as expected. The total line removals and additions looks a lot more than it really is. The additions mainly come from the many yaml files added.

scripts/batch_conv_diags.sh
```diff
@@ -10,9 +10,9 @@
 # PyGSIdir=path/to/PyGSI_Directory
 PyGSIdir=../
 OUTDIR=../
-YAML=$PyGSIdir/test_conv_yaml.yaml
+YAML=$PyGSIdir/yamls/diag_conv_t_ges.yaml
 
 # load environment needed to run python scripts
-source $PyGSIdir/modulefiles/modulefile.PyGSI.hera.bash
+# source $PyGSIdir/modulefiles/modulefile.PyGSI.hera.bash
```
CoryMartin-NOAA, line 16: so is it assume to already have the env on path? do we want to have people use the eva env?

CoryMartin-NOAA (commented):
question, but otherwise looks good

CoryMartin-NOAA (approved):

PR #142, opened 2023-03-17
Bugfix/ensemble da diag read

delippi:
Fixes bug to use Errinv_Input and Obs_Minus_Forecast_unadjusted instead of Errinv_Final and Obs_Minus_Forecast_adjusted respectively. Added logic to use either option. Also fixes a bug introduced by use of numpy.nanmean; some arrays were initialized to nans instead of with zeros to fix this.

### DIFF
--- a/src/pyGSI/ensemble_diags.py
+++ b/src/pyGSI/ensemble_diags.py
@@ -122,9 +122,9 @@ def time_trace(
                 while mem <= n_mem:
                     memid = str(mem).zfill(4)
                     if ob_type == "u" or ob_type == "v":
-                        diagfile = _os.path.join(datapath, f"{expt_name}/{date}/mem{memid}/diag_conv_uv_ges.{date}.nc4")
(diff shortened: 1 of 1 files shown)


src/pyGSI/ensemble_diags.py
```diff
@@ -50,6 +52,8 @@ def time_trace(
       lon_min         : (float) minimum latitude (deg E) for including observation in calculations
       error_max       : (float) maximum error standard deviation for including observation in calculations
       error_min       : (float) minimum error standard deviation for including observation in calculations
+      lomfnbc         : (bool) use non-bias corrected observation minus background. =True: use omf_unadjusted.
```
CoryMartin-NOAA: these args sound very "fortran-namelist-y". What about `use_bc_omf` and `use_input_err`? (note for the first one I suggest false being non bias corrected and true being corrected, that way true is not 'yes use non bias corrected'

delippi: Sounds good to me.

delippi: @CoryMartin-NOAA Done.

CoryMartin-NOAA (approved):

PR #151, opened 2024-12-02
modified scripts/plot_proftrace.py to close figures after saving to file

BrettHoover-NOAA:
Minor changes to scripts/plot_proftrace.py:

1. Close each figure after it has been created and saved to file. In the absence of these `plt.close()` commands `matplotlib.pyplot` will raise a warning if more than 20 figures are opened without being closed. We don't need to keep these open after they have been written to file, so it's fine to include these close commands.
2. Replace checks on input types from `if type(x) != list` to the now standardized `if not isinstance(x, list)` format, for pycodestyle


kevindougherty-noaa (approved):
Looks good to me.

